PR #11, opened 2023-06-14
fix(sdk): properly encode bitcoin and polkadot addresses

acdibble:


niklasnatter (approved):

zoheb391:
can we use this validation on the deposit channel side too? right now we just use some form of regex check i think.

zoheb391 (approved):

PR #12, opened 2023-06-14
feat: funding sdk 

acdibble:


packages/cli/src/main.ts
```diff
@@ -45,9 +51,54 @@ const args = yargs(process.argv.slice(2))
         type: 'string',
         describe:
           'The contract address of the vault when `chainflip-network` is `localnet`',
+      })
+      .option('eth-network', {
+        type: 'string',
+        describe:
+          'The eth network URL to use when `chainflip-network` is `localnet`',
```
niklasnatter, line 58: i think this will be used anyway if it is set. should we ignore this if the `chainflip-network` is not `localnet`?

packages/cli/src/main.ts
```diff
@@ -45,9 +51,54 @@ const args = yargs(process.argv.slice(2))
...
+      })
+      .option('flip-token-contract-address', {
+        type: 'string',
+        describe:
+          'The contract address for the FLIP token when `chainflip-network` is `localnet`',
+      })
+      .option('eth-network', {
+        type: 'string',
+        describe:
+          'The eth network URL to use when `chainflip-network` is `localnet`',
```
niklasnatter, line 94: same as above

packages/shared/src/stateChainGateway/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  );
+
+  const transaction = await stateChainGateway.fundStateChainAccount(
+    accountId,
+    amount,
+  );
+
+  const receipt = await transaction.wait(1);
+
+  assert(receipt.status !== 0, 'Transaction failed');
```
niklasnatter, line 66: the `wait` method will already check if the `status` is 0 and throw an error if it is not: https://github.com/ethers-io/ethers.js/blob/f97b92bbb1bde22fcc44100af78d7f31602863ab/packages/providers/src.ts/base-provider.ts#L1548-L1554

packages/shared/src/stateChainGateway/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    amount,
+    options.signer,
+  );
+
+  const transaction = await stateChainGateway.fundStateChainAccount(
+    accountId,
+    amount,
+  );
+
+  const receipt = await transaction.wait(1);
```
niklasnatter, line 64: `1` would be the default value here anyway, but its fine for me to be explicit

packages/cli/src/main.ts
niklasnatter: not sure how the other arguments like `vault-contract-address` are handled

acdibble: Was going to leave this because the CLI is mostly intended for internal usage but I fixed it up.

PR #15, opened 2023-06-16
feat: Use string instead of number to identify chains

niklasnatter:


packages/sdk/src/__tests__/sdk.test.ts
```diff
@@ -28,40 +28,34 @@ describe(SwapSDK, () => {
     });
 
     it.each([
-      [ChainId.Ethereum, [bitcoin, polkadot]],
-      [ChainId.Polkadot, [ethereum, bitcoin]],
-      [ChainId.Bitcoin, [ethereum, polkadot]],
+      [Chains.Ethereum, [bitcoin, polkadot]],
+      ['Ethereum' as const, [bitcoin, polkadot]],
```
niklasnatter, line 32: make sure that we also accept string version here

acdibble (approved):

PR #17, opened 2023-06-19
chore: Rename token to asset and add Assets enum [WEB-235]

niklasnatter:


packages/cli/src/__tests__/cliExecuteSwap.test.ts
```diff
@@ -4,7 +4,7 @@ import { parseArgs } from '../utils';
 const localnet = `swap
   --wallet-private-key 0x2
   --chainflip-network localnet
-  --dest-token USDC
+  --dest-asset USDC
```
niklasnatter, line 7: will check what we use in the bouncer and create an update pr once this is merged

niklasnatter: will also try to update the docs

zoheb391 (approved):

acdibble (approved):

PR #20, opened 2023-06-20
fix(swap): clean up event listeners

acdibble:


niklasnatter (approved):

zoheb391:
only downside now is that we have to ctrl-c  3x to kill the server. We should prolly abort all ws connections on sigterm, but yeah that can be for a future PR. i'm okay to merge this.

zoheb391 (approved):

niklasnatter (approved):

PR #28, opened 2023-07-03
feat: partnernet [WEB-276]

acdibble:


zoheb391 (approved):

packages/sdk/src/swap/types.ts
```diff
@@ -35,9 +35,7 @@ export interface QuoteResponse extends Route {
   quote: QuoteQueryResponse;
 }
 
-export interface SwapRequest extends Omit<QuoteResponse, 'quote'> {
-  expectedDepositAmount: string;
-}
+export type SwapRequest = QuoteRequest;
```
niklasnatter, line 38: i think it would make sense to move the `destAddress` to the `SwapRequest` type

niklasnatter: we already "make up" a destination address in the swap app because of this. i think it will be the same for others that are trying to use it

niklasnatter (approved):

PR #36, opened 2023-07-03
fix: Include chains and assets in requestDepositAddress return value

niklasnatter:


packages/sdk/src/swap/services/ApiService.ts
```diff
@@ -102,26 +103,29 @@ const getQuote: BackendQuery<QuoteRequest, QuoteResponse> = async (
...
   };
 
   const url = new URL('/swaps', baseUrl).toString();
 
-  const { data } = await axios.post<SwapResponse>(url, body, { signal });
+  const { data } = await axios.post<PostSwapResponse>(url, body, { signal });
 
-  return data;
+  return {
+    ...swapRequest,
```
niklasnatter, line 125: this makes the behaviour consistent to the `getQuote` method

packages/sdk/src/swap/services/ApiService.ts
```diff
@@ -102,26 +103,29 @@ const getQuote: BackendQuery<QuoteRequest, QuoteResponse> = async (
...
+    depositChannelId: data.id,
```
niklasnatter, line 126: this is a breaking change. but i think its important to make it explicit what kind of id this is

packages/sdk/src/swap/sdk.ts
```diff
@@ -53,9 +53,9 @@ export class SwapSDK {
   }
 
   requestDepositAddress(
-    swapRequest: SwapRequest,
+    swapRequest: DepositAddressRequest,
```
zoheb391, line 56: make this depositAddressRequest too?

acdibble (approved):

PR #38, opened 2023-07-05
feat: Add executeCall method to swap sdk and cli [WEB-272]

niklasnatter:


packages/cli/src/cliExecuteCall.ts
```diff
@@ -0,0 +1,86 @@
...
+    z.union([
+      z.object({
+        chainflipNetwork: z.literal('localnet'),
+        ethNetwork: z.string(),
+        srcTokenContractAddress: z.string(),
+        vaultContractAddress: z.string(),
+      }),
+      z.object({
+        chainflipNetwork,
+        ethNetwork: z
```
niklasnatter, line 28: cumbersome workaround to satisfy typescript when destructing this below 😕 
```
  const {
    chainflipNetwork: network,
    vaultContractAddress,
    srcTokenContractAddress,
    ...callParams
  } = args;
```

packages/cli/src/cliExecuteCall.ts
```diff
@@ -0,0 +1,86 @@
...
+    process.env.ALCHEMY_KEY
+      ? new providers.AlchemyProvider(ethNetwork, process.env.ALCHEMY_KEY)
+      : getDefaultProvider(ethNetwork),
+  );
+
+  const {
+    chainflipNetwork: network,
+    vaultContractAddress,
+    srcTokenContractAddress,
+    ...callParams
```
niklasnatter, line 70: if we just pass `args`, the `executeSwap` method will parse the object and complain that is has additional keys (`chainflipNetwork`)

packages/shared/src/enums.ts
```diff
@@ -42,8 +42,26 @@ export const assetDecimals = {
   [Assets.BTC]: 8,
 } satisfies Record<Asset, number>;
 
+// https://github.com/chainflip-io/chainflip-backend/blob/a2a3c2e447e7b629c4b96797d9eed22eb5b87a0b/state-chain/primitives/src/chains/assets.rs#L51-L59
+export const assetContractIds: Record<Asset, number> = {
```
niklasnatter, line 46: https://linear.app/chainflip/issue/WEB-283/export-constants-for-partners-use

packages/shared/src/vault/schemas.ts
```diff
@@ -71,3 +74,50 @@ export const executeSwapParamsSchema = z.union([
...
+export type TokenCallParams = z.infer<typeof tokenCallParamsSchema>;
+
+export const executeCallParamsSchema = z.union([
+  nativeCallParamsSchema,
+  tokenCallParamsSchema,
+]);
+
+export type ExecuteCallParams = z.infer<typeof executeCallParamsSchema>;
+
+export const executeOptionsSchema = z.intersection(
```
niklasnatter, line 108: feels a bit hard/cumbersome to get the validation right this way and the error messages are not too helpful. think its okay for now, but we might want to validate the possible combinations and stuff in code when we get more combinations

packages/cli/src/cliExecuteCall.ts
```diff
@@ -0,0 +1,83 @@
...
+} from '@/shared/parsers';
+import {
+  type ExecuteOptions,
+  executeCall,
+  type ExecuteCallParams,
+} from '@/shared/vault';
+import { askForPrivateKey, getEthNetwork } from './utils';
+
+export const schema = z
+  .intersection(
```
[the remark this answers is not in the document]
niklasnatter: might look into reusing the yarg object here and use `zod` only in the `executeCall` in a separate pr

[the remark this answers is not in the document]
niklasnatter: https://github.com/chainflip-io/chainflip-sdk-monorepo/pull/46

packages/cli/src/utils.ts
```diff
@@ -74,6 +74,62 @@ export const parseArgs = (args: string[]) =>
             'The eth network URL to use when `chainflip-network` is `localnet`',
         });
     })
+    .command('call', '', (y) => {
+      y.option('src-asset', {
+        choices: Object.values(Assets),
+        // demandOption: true,
```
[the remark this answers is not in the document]
niklasnatter: copied from the swap command, but will change in https://github.com/chainflip-io/chainflip-sdk-monorepo/pull/46

packages/shared/src/guards.ts
```diff
@@ -14,3 +21,11 @@ export function assert(condition: unknown, message: string): asserts condition {
...
   throw error;
 }
+
+export const isTokenSwap = (
+  params: ExecuteSwapParams,
+): params is TokenSwapParams => 'srcAsset' in params;
+
+export const isTokenCall = (
+  params: ExecuteCallParams,
+): params is TokenCallParams => 'srcAsset' in params;
```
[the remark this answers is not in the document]
niklasnatter: not sure because i need to change the typeguard `params is TokenCallParams`

packages/cli/src/cliExecuteCall.ts
```diff
@@ -0,0 +1,92 @@
...
+    }),
+  )
+  .transform(({ destAsset, srcAsset, ...rest }) => ({
+    ...rest,
+    destChain: assetChains[destAsset],
+    destAsset,
+    ...(srcAsset && { srcAsset }),
+  }));
+
+export default async function cliExecuteCall(unvalidatedArgs: unknown) {
```
zoheb391, line 59: so chainflip-backend calls this `executeSwapAndCall`. This does the same right? Why don't we use the same name?

niklasnatter: i used the naming of the contract because that is what is invoked: https://github.com/chainflip-io/chainflip-eth-contracts/blob/cd289f4b0f78a7a9ea55bc221e88bba528ee9ff4/contracts/Vault.sol#L347

niklasnatter: where did you find the `executeSwapAndCall`?

PR #40, opened 2023-07-06
feat(swap): lookup swap by tx hash [WEB-240]

acdibble:
need to publish the SDK + update the swapping app


zoheb391 (approved):

niklasnatter (approved):
nice! need to adjust the docs when merged

PR #41, opened 2023-07-07
chore: partnernet trigger

tomjohnburton:


.github/workflows/test-swap-service-partnernet.yml
```diff
@@ -0,0 +1,76 @@
+name: Test Swap Service
```
niklasnatter, line 1: whats the difference to the default test workflow?

PR #42, opened 2023-07-07
feat(sdk, swap): add ccm metadata params [WEB-271]

zoheb391:
closes WEB-271

### DIFF
--- a/packages/shared/src/parsers.ts
+++ b/packages/shared/src/parsers.ts
@@ -18,7 +18,9 @@ const errorMap: ZodErrorMap = (issue, context) => ({
 export const string = z.string({ errorMap });
 export const number = z.number({ errorMap });
 export const numericString = string.regex(/^[0-9]+$/);
-export const hexString = string.regex(/^0x[0-9a-f]+$/i);
+export const hexString = z.custom<`0x${string}`>((val) =>
+  /^0x[0-9a-f]+$/i.test(val as string),
+);
 export const hexStringFromNumber = numericString
   .transform((arg) => ethers.BigNumber.from(arg).toHexString())
   .refine((arg) => arg.startsWith('0x'));
--- a/packages/shared/src/parsers.ts
+++ b/packages/shared/src/parsers.ts
@@ -18,8 +18,8 @@ const errorMap: ZodErrorMap = (issue, context) => ({
 export const string = z.string({ errorMap });
 export const number = z.number({ errorMap });
 export const numericString = string.regex(/^[0-9]+$/);
-export const hexString = z.custom<`0x${string}`>((val) =>
-  /^0x[0-9a-f]+$/i.test(val as string),
+export const hexString = z.custom<`0x${string}`>(
+  (val) => typeof val === 'string' && /^0x[0-9a-f]+$/i.test(val),
 );
 export const hexStringFromNumber = numericString
   .transform((arg) => ethers.BigNumber.from(arg).toHexString())
--- a/packages/shared/src/schemas.ts
+++ b/packages/shared/src/schemas.ts
@@ -12,7 +12,7 @@ export type QuoteQueryParams = z.infer<typeof quoteQuerySchema>;
 
 export type ByteString = string | `0x${string}`;
 export interface CcmMetadata {
-  gas_budget: string | number;
+  gas_budget: `0x${string}` | number;
   message: ByteString;
   source_address: string;
   source_chain: 'Bitcoin' | 'Ethereum' | 'Polkadot';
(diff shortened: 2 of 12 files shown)


packages/shared/src/schemas.ts
```diff
@@ -10,12 +10,34 @@ export const quoteQuerySchema = z.object({
...
+  message: ByteString;
+  source_address: string;
+  source_chain: 'Bitcoin' | 'Ethereum' | 'Polkadot';
+  cf_parameters?: ByteString;
+}
+
+export const ccmMetadataSchema = z.object({
+  gas_budget: z.union([numericString, z.number()]),
+  message: z.union([hexString, z.string()]),
+  source_address: z.string(),
```
niklasnatter, line 25: what kind of address is this?

packages/shared/src/schemas.ts
```diff
@@ -10,12 +10,34 @@ export const quoteQuerySchema = z.object({
...
+  source_chain: z.union([
```
niklasnatter, line 26: we can infer this from the `srcAsset`

packages/shared/src/schemas.ts
niklasnatter: if it is the token address, we can infer it from the `srcAsset`

packages/shared/src/schemas.ts
```diff
@@ -10,12 +10,34 @@ export const quoteQuerySchema = z.object({
 
 export type QuoteQueryParams = z.infer<typeof quoteQuerySchema>;
 
+export type ByteString = string | `0x${string}`;
+export interface CcmMetadata {
+  gas_budget: string | number;
```
acdibble, line 15: This might actually be:
```suggestion
  gas_budget: `0x${string}` | number;
```

packages/shared/src/schemas.ts
acdibble: might as well keep it so we can handle arbitrum in the future

niklasnatter: thats fine for me in general, but we should add it to the root of the `postSwapSchema`, bot inside the `ccmMetadata` then

niklasnatter: would also need to add `destChain` for the `destAsset` to keep things consistent

zoheb391: i think we should change the shape of the root object when arbitrum is ready so we can capture all the changes together (broker api, front-end) etc.. Don't want to increase the scope of this ticket.  @acdibble @niklasnatter

packages/shared/src/schemas.ts
zoheb391: @acdibble i'm not sure. how do i find out?

packages/shared/src/schemas.ts
```diff
@@ -10,12 +15,30 @@ export const quoteQuerySchema = z.object({
 
 export type QuoteQueryParams = z.infer<typeof quoteQuerySchema>;
 
+export type ByteString = string | `0x${string}`;
+export interface CcmMetadata {
+  gas_budget: `0x${string}` | number;
+  message: ByteString;
+  source_address: 0; // tb removed
+  cf_parameters?: ByteString;
+}
```
niklasnatter, line 24: not sure if this works

```suggestion
export type CcmMetadata = z.infer<typeof ccmMetadataSchema>;
```

packages/shared/src/parsers.ts
```diff
@@ -18,7 +18,9 @@ const errorMap: ZodErrorMap = (issue, context) => ({
 export const string = z.string({ errorMap });
 export const number = z.number({ errorMap });
 export const numericString = string.regex(/^[0-9]+$/);
-export const hexString = string.regex(/^0x[0-9a-f]+$/i);
+export const hexString = z.custom<`0x${string}`>((val) =>
+  /^0x[0-9a-f]+$/i.test(val as string),
```
acdibble, line 22:
```
/^0x/.test(['0x1234'])
// => true
```

```suggestion
  typeof val === 'string' && /^0x[0-9a-f]+$/i.test(val),
```

packages/swap/src/utils/broker.ts
```diff
@@ -26,8 +48,20 @@ const requestValidators = {
...
+            source_chain: chainflipChain,
+            source_address: z.literal(0),
+          }),
+        )
+        .optional(),
     ])
-    .transform(([a, b, c, d]) => [a, b, c, d]),
+    .transform(([a, b, c, d, e]) =>
+      [a, b, c, d, transformObjToSnakeCase(e)].filter(
+        (item) => item !== undefined && item !== null,
```
acdibble, line 62: Let's add this to the `shared/guards.ts`:
```ts
export const isNotNullish = <T>(value: T | null | undefined): value is T =>
  value !== null && value !== undefined;
```

```suggestion
        isNotNullish,
```

acdibble (approved):

niklasnatter (approved):

PR #44, opened 2023-07-10
fix(sdk,shared): separate approvals from smart contract calls [WEB-299]

acdibble:


packages/sdk/src/swap/sdk.ts
```diff
@@ -72,9 +74,22 @@ export class SwapSDK {
 
   executeSwap(params: ExecuteSwapParams): Promise<ContractReceipt> {
     assert(this.signer, 'No signer provided');
-    return executeSwap(params, {
-      network: this.network,
+    return executeSwap(params, { network: this.network, signer: this.signer });
```
niklasnatter, line 77: would prefer returning the hash here too to make things consistent with the other methods

packages/sdk/src/swap/sdk.ts
```diff
@@ -72,9 +74,22 @@ export class SwapSDK {
...
+  }
+
+  async approveVault(
```
niklasnatter, line 80: should we expose a method like `needsVaultApproval`?

packages/sdk/src/funding/index.ts
```diff
@@ -64,31 +60,22 @@ export class FundingSDK {
...
-    return balance.toBigInt();
+    return getFlipBalance(this.options.network, this.options.signer);
   }
 
   /**
    * @param amount the amount of FLIP to request approval for
    * @returns the transaction hash or null if no approval was required
    */
-  async requestFlipApproval(
+  async approveStateChainGateway(
```
niklasnatter, line 70: how do we determine if we need an approval in the auctions app? should we expose a method like `needsStateChainGatewayApproval`?

packages/shared/src/consts.ts
```diff
@@ -0,0 +1,46 @@
+import { Asset, Chain, Chains, Assets, ChainflipNetworks } from './enums';
+
+// !!!!! IMPORTANT !!!!!
+// Do not change these indices.
+export const chainMap: Record<Chain, number> = {
```
niklasnatter, line 5: `chainContractIds`?

packages/shared/src/contracts.ts
```diff
@@ -53,22 +29,31 @@ export const getStateChainGatewayContractAddress = (
...
+  erc20: ERC20,
+  allowance: BigNumberish,
+  nonce?: bigint | number | string,
+): Promise<ContractReceipt | null> => {
+  const amountBigNumber = BigNumber.from(amount);
+  const allowanceBigNumber = BigNumber.from(allowance);
+  if (allowanceBigNumber.gte(amountBigNumber)) return null;
+  const requiredAmount = amountBigNumber.sub(allowanceBigNumber);
+  const tx = await erc20.approve(spenderAddress, requiredAmount, { nonce });
+  return tx.wait(1);
```
niklasnatter, line 56: think we should try to handle transactions consistent - either we should always wait for inclusion before returning or return immediately after submitting

acdibble: I don't know anywhere we don't wait for inclusion before returning a hash to the consumer

packages/shared/src/vault/approval.ts
```diff
@@ -0,0 +1,59 @@
...
+  return checkAllowance(
+    params.amount,
+    vaultContractAddress,
+    erc20Address,
+    opts.signer,
+  );
+};
+
+export const approveVault = async (
+  params: TokenSwapParams,
```
albert-llimos, line 37: Does it make sense to have this TokenSwapParams here? That contains destination parameters which won't be used here. Might be weird/misleading in the SDK.

acdibble: Fixed

packages/sdk/src/swap/sdk.ts
zoheb391: +1 but could be its own ticket.

zoheb391 (approved):

albert-llimos (approved):
I tested it with the bouncer and it fixed the issue :rocket:

PR #46, opened 2023-07-10
feat: Use yargs option type to type command arguments

niklasnatter:


packages/cli/src/main.ts
```diff
@@ -1,18 +1,25 @@
...
-  cliExecuteSwap(args);
-} else if (command === 'call') {
-  cliExecuteCall(args);
-} else if (command === 'fund-state-chain-account') {
-  cliFundStateChainAccount(args);
-}
+yargs(process.argv.slice(2))
+  .scriptName('chainflip-cli')
+  .usage('$0 <cmd> [args]')
+  .command('swap', '', cliExecuteSwapOptions, cliExecuteSwap)
```
niklasnatter, line 16: we can pass a handler here that will be called with the parsed args: http://yargs.js.org/docs/#api-reference-commandcmd-desc-builder-handler

zoheb391 (approved):
neat

PR #49, opened 2023-07-18
refactor: expose broker client in CLI lib

acdibble:
Instead of refactoring the code to be js environment agnostic, I moved the code that use Node.js APIs to a separate directory and then restricted its import in the SDK via eslint


packages/cli/src/commands/cliRequestSwapDepositAddress.ts
```diff
@@ -0,0 +1,55 @@
...
+    demandOption: true,
+  },
+} as const satisfies { [key: string]: Options };
+
+export default async function cliRequestSwapDepositAddress(
+  args: ArgumentsCamelCase<InferredOptionTypes<typeof yargsOptions>>,
+) {
+  const client = await BrokerClient.create({ url: args.brokerUrl });
+
+  const result = await client.requestSwapDepositAddress({
```
niklasnatter, line 42: could add a test cases if this is called with the correct params

zoheb391 (approved):

PR #52, opened 2023-07-18
feat: Make deposit channel id deterministic by using native channel id instead of uuid

niklasnatter:


zoheb391 (approved):

packages/swap/src/routes/swap/index.ts
```diff
@@ -35,7 +35,8 @@ type SwapWithBroadcast = Swap & {
     | null;
 };
 
-const uuidRegex = /^[a-f\d]{8}(-[a-f\d]{4}){4}[a-f\d]{8}$/i;
+const channelIdRegex = /^(\d+)-([a-z]+)-([a-z]+)$/i;
```
acdibble, line 38: channel id should be a number
```suggestion
const channelIdRegex = /^(\d+)-([a-z]+)-(\d+)$/i;
```

acdibble, line 38: or we can use named capture groups to make it a bit clearer
```suggestion
const channelIdRegex = /^(?<blockId>\d+)-(?<chain>[a-z]+)-(?<channelId>\d+)$/i;
```

```ts
/^(?<blockId>\d+)-(?<chain>[a-z]+)-(?<channelId>\d+)$/i.exec('1234-ETH-2').groups
// => {blockId: '1234', chain: 'ETH', channelId: '2'}
```

niklasnatter: thats nice!

packages/sdk/src/swap/types.ts
```diff
@@ -50,21 +50,23 @@ export interface SwapStatusRequest {
 
 export interface CommonStatusFields extends ChainsAndAssets {
   destAddress: string;
-  depositAddress: string;
-  expectedDepositAmount: string;
+  depositAddress: string | undefined;
```
niklasnatter, line 53: `undefined` if swap was triggered via smart contract call

PR #53, opened 2023-07-19
refactor(shared): make smart contract params more consistent

acdibble:


zoheb391 (approved):

niklasnatter (approved):

packages/shared/src/guards.ts
```diff
@@ -27,8 +28,8 @@ export function assert(condition: unknown, message: string): asserts condition {
 
 export const isTokenSwap = (
   params: ExecuteSwapParams,
-): params is TokenSwapParams => 'srcAsset' in params;
+): params is TokenSwapParams => params.srcAsset !== Assets.ETH;
```
niklasnatter, line 31: im afraid this changes the bahaviour in the case where `srcAsset` is not set/undefined

niklasnatter: it should still be a token swap/call if `srcAsset` is not set

PR #57, opened 2023-07-19
chore(cli): add require

acdibble:
oof commonjs v modules

### DIFF
--- a/packages/cli/tsup.config.mjs
+++ b/packages/cli/tsup.config.mjs
@@ -5,10 +5,20 @@ export default defineConfig({
   treeshake: true,
   minify: false,
   dts: true,
-  format: ['cjs', 'esm'],
+  format: 'esm',
   entry: {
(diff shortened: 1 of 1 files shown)


packages/cli/tsup.config.mjs
```diff
@@ -5,10 +5,20 @@ export default defineConfig({
...
-  splitting: true,
+  sourcemap: true,
+  target: 'es2022',
+  banner: {
+    js: `
+import { fileURLToPath } from 'url';
+import { createRequire as topLevelCreateRequire } from 'module';
+const require = topLevelCreateRequire(import.meta.url);
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
niklasnatter, line 21: do we use them in the sdk somewhere?

niklasnatter (approved):

PR #65, opened 2023-07-21
feat(swap): add CCM event handler [WEB-337]

acdibble:


packages/swap/prisma/schema.prisma
```diff
@@ -71,7 +71,8 @@ model Swap {
   swapDepositChannel        SwapDepositChannel? @relation(fields: [swapDepositChannelId], references: [id])
   swapDepositChannelId      BigInt?
   txHash                    String?
-  egress                    Egress?
+  egress                    Egress?             @relation(fields: [egressId], references: [id])
+  egressId                  BigInt?
```
acdibble, line 75: The CCM swaps share an egress ID

packages/swap/src/event-handlers/networkEgressScheduled.ts
```diff
@@ -22,13 +22,20 @@ export default async function networkEgressScheduled({
 }: EventHandlerArgs): Promise<void> {
   const { id, asset, amount } = eventArgs.parse(event.args);
 
-  await prisma.egress.create({
-    data: {
+  await prisma.egress.upsert({
```
niklasnatter, line 25: there will be two egress scheduled events?

niklasnatter (approved):

packages/swap/src/event-handlers/networkEgressScheduled.ts
acdibble: No there will be 1 egress for 2 swaps

PR #75, opened 2023-07-26
fix: Support swap from ethereum to ethereum

niklasnatter:


packages/sdk/src/swap/services/ApiService.ts
```diff
@@ -30,7 +30,7 @@ const getPossibleDestinationChains = async (
   network: ChainflipNetwork,
 ): Promise<ChainData[]> => {
   if (sourceChain === Chains.Ethereum)
-    return [bitcoin(network), polkadot(network)];
+    return [ethereum(network), bitcoin(network), polkadot(network)];
```
niklasnatter, line 33: it might not be the fastest option, but we still support swaps from ethereum to ethereum like flip->usdc

acdibble (approved):

zoheb391 (approved):

PR #78, opened 2023-07-28
chore(swap): add createdAt to status

zoheb391:


packages/swap/src/routes/swap.ts
```diff
@@ -129,6 +129,7 @@ router.get(
       srcAsset,
       destAsset,
       destAddress: readField('destAddress'),
+      createdAt: swapDepositChannel?.createdAt.valueOf(),
```
niklasnatter, line 132: might be a bit confusing as its not defined for smart contract calls
```suggestion
      depositAddressCreatedAt: swapDepositChannel?.createdAt.valueOf(),
```

niklasnatter (commented):
need to add this to the sdk types to be able to use it

packages/swap/src/routes/swap.ts
zoheb391: good point 🤦

packages/sdk/src/swap/types.ts
```diff
@@ -62,13 +62,15 @@ export type SwapStatusResponse = CommonStatusFields &
         swapId: string;
         depositAmount: string;
         depositReceivedAt: number;
+        depositAddressCreatedAt: number;
```
niklasnatter, line 65: think you can move it to `CommonStatusFields` 🙂

niklasnatter: and it needs to be `number | undefined` i guess

niklasnatter: its not great that we dont check if the response matches the types we have anywhere 🙈

acdibble (approved):

niklasnatter (approved):

PR #80, opened 2023-07-31
fix(sdk): working directory check to tag script

zoheb391:


packages/shared/src/vault/schemas.ts
```diff
@@ -85,21 +85,18 @@ const nativeCallParamsSchema = ethereumBase.extend({
...
-const ccmMetadataSchema = z.object({
-  message: hexString,
-  gasBudget: numericString,
-});
-
 const tokenCallParamsSchema = z.union([
   ethereumBase.extend({
     srcAsset: z.literal(Assets.FLIP),
     destAsset: z.union([z.literal(Assets.USDC), z.literal(Assets.ETH)]),
-    ccmMetadata: ccmMetadataSchema,
```
niklasnatter, line 97: 🤔

zoheb391: this was published because we don't check that the working directory is clean.

acdibble (approved):

niklasnatter (approved):

PR #81, opened 2023-08-01
chore(sdk): merge execute swap and call

zoheb391:
closes WEB-341


packages/cli/src/commands/cliExecuteSwap.ts
```diff
@@ -86,6 +94,11 @@ export default async function cliExecuteSwap(
       destAsset: args.destAsset,
       amount: args.amount,
       destAddress: args.destAddress,
+      ccmMetadata: args.message &&
```
niklasnatter, line 97: maybe throw an error if one of them is missing like we do in the `cliRequestSwapDepositAddress`

niklasnatter (approved):
looks good to me otherwise! need to update the docs for this

PR #82, opened 2023-08-02
refactor(shared): rename schemas

acdibble:


packages/shared/src/vault/schemas.ts
```diff
@@ -17,78 +17,81 @@ const bytesToHex = (arr: Uint8Array | number[]) =>
 
 const utf8ToHex = (str: string) => `0x${Buffer.from(str).toString('hex')}`;
 
-const base = z.object({
+const fromEth = z.object({
```
zoheb391, line 20: i think the convention we have been following is using the fullname of the chain and short-hand for the asset. I think that will make things easier. 
could maybe drop camelCase here as well to make it more readable and use snake case without words `ethereum_flip`  = ethereum to flip.

acdibble: I removed `from` from the beginning and that increased readability.

The convention is followed here where possible, e.g. for asset to asset `ethToDot`, `ethToBtc`, but had to introduce some new ways, like for asset to network `ethToEthereum` (sets the `destChain` but not the `destAsset`), or asset to any token on the network `flipToEthereumAsset`, or `erc20ToBtc` for token swaps etc.

zoheb391 (approved):

PR #137, opened 2023-08-15
feat(sdk,swap): add arbitrum

zoheb391:
on request from @martin-chainflip


packages/sdk/src/swap/assets.ts
```diff
@@ -68,3 +68,27 @@ export const btc$: (network: ChainflipNetwork) => AssetData = (network) => ({
...
   minimumSwapAmount: getMinimumSwapAmount(network, Assets.BTC),
 });
+
+export const arbeth$: (network: ChainflipNetwork) => AssetData = (network) => ({
+  id: Assets.ARBETH,
+  chain: Chains.Arbitrum,
+  contractAddress: undefined,
+  decimals: assetDecimals[Assets.ARBETH],
+  name: 'Arbitrum',
+  symbol: 'ARBETH',
```
niklasnatter, line 78: think i woul use `ETH` and `USDC` as symbol

niklasnatter: but we can still change it later

packages/sdk/src/swap/assets.ts
```diff
@@ -68,3 +68,27 @@ export const btc$: (network: ChainflipNetwork) => AssetData = (network) => ({
...
+  isMainnet: !isTestnet(network),
+  minimumDepositAmount: getMinimumDepositAmount(network, Assets.ARBETH),
+  minimumSwapAmount: getMinimumSwapAmount(network, Assets.ARBETH),
+});
+
+export const arbusd$: (network: ChainflipNetwork) => AssetData = (network) => ({
+  id: Assets.ARBUSDC,
+  chain: Chains.Arbitrum,
+  contractAddress: undefined,
```
niklasnatter, line 87: this should have a contract address i think

packages/shared/src/consts.ts
```diff
@@ -27,6 +29,8 @@ const MINIMUM_SWAP_AMOUNTS: Partial<
     [Assets.USDC]: '1000000',
     [Assets.BTC]: '390000',
     [Assets.DOT]: '2000000000',
+    [Assets.ARBETH]: '580000000000000', // TODO: ??
```
niklasnatter, line 32: i got the value by querying the partnernet rpc. but this is not deployed to partnernet, right?

packages/sdk/src/swap/services/ApiService.ts
```diff
@@ -23,18 +23,31 @@ const getChains = async (network: ChainflipNetwork): Promise<ChainData[]> => [
...
   sourceChain: Chain,
   network: ChainflipNetwork,
 ): Promise<ChainData[]> => {
   if (sourceChain === Chains.Ethereum)
-    return [ethereum(network), bitcoin(network), polkadot(network)];
+    return [
+      ethereum(network),
+      bitcoin(network),
+      polkadot(network),
+      arbitrum(network),
```
niklasnatter, line 38: maybe include this only for `backspin` for now? otherwise partners will already get the asset/chain even if it snot supported by `partnernet`

packages/sdk/src/swap/assets.ts
acdibble: Agreed, it's similar to our `swapToken`/`callToken` stuff. It makes sense with the context we have but for users of the SDK without the context, it seems like an arbitrary decision.

Not a fan of this convention in general because it overloads the asset id to be a compound identifier of chain and asset ID but only for non-Ethereum EVM chains.

acdibble (approved):

packages/sdk/src/swap/assets.ts
zoheb391: we can decide on this later.. for now especially during dev it will help to have it be called ARBETH especially since that's what they're calling it on the backend. And calling this ETH would mean a bigger refactor of how we define Assets and I don't think that should be a part of this PR

packages/shared/src/consts.ts
zoheb391: no its not even on main yet.

PR #194, opened 2023-09-04
chore: Split up network options and transaction options

niklasnatter:


packages/shared/src/stateChainGateway/__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -4,13 +4,7 @@ exports[`executeRedemption executes the redemption 1`] = `
 [
   "0x1234",
   {
-    "network": "sisyphos",
     "nonce": 1,
-    "signer": VoidSigner {
```
niklasnatter, line 9: this was the initial reason for the pr - we pass the `signer` and `network` to `ethers.js` at the moment. this causes ethers to throw an error that it does not expect a `signer` property

packages/shared/src/vault/executeSwap.ts
```diff
@@ -106,43 +101,38 @@ const callNative = async (
...
 const callToken = async (
   params: TokenCallParams,
-  {
-    signer,
-    network,
-    vaultContractAddress: vaultAddress,
-    srcTokenContractAddress: tokenAddress,
-    ...opts
-  }: ExecuteOptions,
+  opts: SwapNetworkOptions,
```
zoheb391, line 112: i think i would rename opts to `networkOpts` everywhere we have two types of opts for the sake of clarity

zoheb391 (approved):
the documentation changes should go to the new doc repo. (push to main) instead of the old one.

packages/shared/src/vault/executeSwap.ts
niklasnatter: sounds good!

PR #200, opened 2023-09-08
fix: Use safe-stable-stringify to stringify log context

niklasnatter:


packages/swap/package.json
```diff
@@ -63,6 +63,7 @@
     "pg": "^8.11.3",
     "qs": "^6.11.2",
     "rxjs": "^7.8.1",
+    "safe-stable-stringify": "^2.4.3",
```
niklasnatter, line 66: this prevents problems when serializing circular objects or bigint instances. it is the same library that is used by our logging library `winston`: https://github.com/winstonjs/logform#json

PR #220, opened 2023-09-18
feat(sdk): Add getPendingRedemption method to FundingSDK

niklasnatter:


packages/shared/src/stateChainGateway/index.ts
```diff
@@ -79,4 +86,17 @@ export const getRedemptionDelay = (
   return stateChainGateway.REDEMPTION_DELAY();
 };
 
+export const getPendingRedemption = async (
```
niklasnatter, line 89: TODO: add test case

packages/sdk/src/funding/sdk.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  async getPendingRedemption(
+    accountId: `0x${string}`,
+  ): Promise<PendingRedemption | undefined> {
+    return getPendingRedemption(accountId, this.options);
+  }
+
+  /**
+   * @param amount the amount of FLIP to request approval for
```
zoheb391, line 79:
```suggestion
   * @param the amount of FLIP to request approval for
```

packages/shared/src/stateChainGateway/index.ts
```diff
@@ -79,4 +86,17 @@ export const getRedemptionDelay = (
...
+  networkOpts: FundingNetworkOptions,
+): Promise<PendingRedemption | undefined> => {
+  const stateChainGateway = getStateChainGateway(networkOpts);
+  const pendingRedemption =
+    await stateChainGateway.getPendingRedemption(accountId);
+
+  return pendingRedemption.amount > 0
+    ? stateChainGateway.getPendingRedemption(accountId)
+    : undefined;
+};
```
zoheb391, line 100: not clear why we're doing it like this?

niklasnatter: do you mean the `pendingRedemption.amount > 0`? thats because there are no `null` values in solidity. if something is not set, it has the initial value for the type: https://www.wtf.academy/en/solidity-start/InitialValue/

packages/sdk/src/funding/index.ts
```diff
@@ -1,86 +1,2 @@
...
-    const receipt = await approveStateChainGateway(
-      amount,
-      this.options,
-      txOpts,
-    );
-
-    return receipt ? (receipt.hash as `0x${string}`) : null;
-  }
-}
+export { FundingSDK, type FundingSDKOption } from './sdk';
```
niklasnatter, line 1: makes it easier to understand what is exported from the package (als makes things consistent with `SwapSDK`)

packages/sdk/src/funding/sdk.ts
acdibble, line 1: I think it's fine to leave this in the index.ts file and include more exports there

acdibble (approved):

zoheb391 (approved):

PR #224, opened 2023-09-19
fix(sdk): Pass valid BytesLike string to ethers calls

niklasnatter:


packages/shared/src/vault/executeSwap.ts
```diff
@@ -38,7 +38,7 @@ const swapNative = async (
     chainContractIds[destChain],
     destAddress,
     assetContractIds[destAsset],
-    '',
+    '0x',
```
niklasnatter, line 41: the alternative would be `new Uint8Array()`

packages/shared/src/schemas.ts
```diff
@@ -49,8 +49,8 @@ export const quoteResponseSchema = z.union([
   z
     .object({
       id: z.string(),
-      intermediate_amount: z.string(),
-      egress_amount: z.string(),
+      intermediate_amount: numericString,
```
niklasnatter, line 52: make type more descriptive

packages/shared/src/schemas.ts
```diff
@@ -16,8 +16,8 @@ export const quoteQuerySchema = z.object({
 export type QuoteQueryParams = z.infer<typeof quoteQuerySchema>;
 
 export const ccmMetadataSchema = z.object({
-  gasBudget: z.union([hexString, z.number()]),
-  message: z.union([hexString, z.string()]),
+  gasBudget: numericString,
```
niklasnatter, line 19: dont think there is a use case to accept a hex string here

acdibble (approved):

PR #225, opened 2023-09-19
fix: Pass gas budget to broker as hex encoded number

niklasnatter:


zoheb391 (approved):

zoheb391 (changes requested):
wait.. don't we already apply the camel case to snake transformer on the request objects.

niklasnatter:
> wait.. don't we already apply the camel case to snake transformer on the request objects?

yes we do. the `cfParameters` will be transformed to `cf_parameters` before it is passed to the broker

PR #234, opened 2023-09-29
feat: Include fees in quote from SDK [WEB-388]

niklasnatter:


acdibble (approved):

packages/swap/prisma/schema.prisma
```diff
@@ -165,3 +165,13 @@ model ThirdPartySwap {
 
   @@schema("public")
 }
+
+model Pool {
+  id               Int   @id @default(autoincrement())
+  baseAsset        Asset
+  pairAsset        Asset
+  feeHundredthPips Int
```
zoheb391, line 173: should we call this liquidityFee or is that implied ?

packages/swap/src/quoting/quotes.ts
```diff
@@ -83,3 +132,126 @@ export const buildQuoteRequest = (query: QuoteQueryParams): QuoteRequest => {
     deposit_amount: amount,
   };
 };
+
+export const calculateIncludedFees = (
+  request: QuoteRequest,
+  quote: MarketMakerResponse,
```
zoheb391, line 138: The `MarketMakerResponse` can technically also be a `BrokerQuoteResponse` too right? are we just calling them both `MarketMakerResponse` ?

packages/swap/prisma/schema.prisma
niklasnatter: i called it `feeHundredthPips` because that is how it is called in the protocol, but will change it to `liquidityFeeHundredthPips`

packages/swap/src/quoting/quotes.ts
niklasnatter: yes unfortunately we have a single type only for both of them at the moment. i will add a `BrokerQuoteResponse` type to make this explicit

zoheb391 (approved):

PR #241, opened 2023-10-05
fix: Always use last deposit channel

niklasnatter:


packages/swap/src/event-handlers/swapScheduled.ts
```diff
@@ -58,25 +57,22 @@ export default async function swapScheduled({
...
+      orderBy: { id: 'desc' },
     });
-    if (channels.length === 0) {
+    if (!channel) {
       logger.info(
         `SwapScheduled: SwapDepositChannel not found for depositAddress ${depositAddress}`,
       );
       return;
     }
-    assert(
```
niklasnatter, line 74: we did not think about this case when removing the `expiryBlock` in https://github.com/chainflip-io/chainflip-sdk-monorepo/pull/227

niklasnatter: after removing the `expiryBlock`, we cannot ensure that we have a single active channel in the database anymore because we dont know if a channel is expired. for now, we just assume that the latest channel is active

zoheb391 (approved):

PR #250, opened 2023-10-11
feat(swap): channel expiry logic [WEB-496]

acdibble:


packages/swap/prisma/schema.prisma
```diff
@@ -36,16 +36,19 @@ enum Asset {
...
   srcChain              Chain
   srcAsset              Asset
   depositAddress        String
-  expectedDepositAmount Decimal  @db.Decimal(30, 0)
+  expectedDepositAmount Decimal      @db.Decimal(30, 0)
   destAsset             Asset
   destAddress           String
   issuedBlock           Int
-  createdAt             DateTime @default(now())
+  expiryBlock           Int?
```
zoheb391, line 48: wanna leave a comment here to delete this column later?

packages/swap/src/processBlocks.ts
```diff
@@ -53,7 +53,12 @@ const fetchBlocks = async (height: number): Promise<Block[]> => {
   throw new Error('failed to fetch batch');
 };
 
-export default async function processBlocks() {
+export default async function processBlocks(
+  preBlockHook?: (
```
zoheb391, line 57: why are we passing this as an argument? should be able to just import this right?

jerryafr (approved):

zoheb391 (approved):

packages/swap/src/processBlocks.ts
acdibble: was just imagining the inevitable scenario of swapping app ephemeral processors. i've inlined it for now

PR #251, opened 2023-10-13
feat: allow for broker configuration in sdk [WEB-508]

acdibble:


packages/swap/src/handlers/openSwapDepositChannel.ts
acdibble, line 1: Removed this logic from the existing express route handler. Now it's used in the express route handler and the tRPC handler

jerryafr (approved):

packages/swap/src/routes/swap.ts
```diff
@@ -176,70 +172,18 @@ router.get(
   }),
 );
 
-let client: BrokerClient | undefined;
-
+// TODO(major): remove this handler. it's replaced by tRPC
```
zoheb391, line 175: any reason why we didn't remove this ? the sdk directly calls the trpc method now right?

zoheb391 (approved):
looks great. i think its works really well for our use case. Do you see any downside of getting rid of our rest endpoints though? does this mean the sdk is the only way to interact with the swap backend? not sure what the trpc experience is when used client side.

acdibble:
> looks great. i think its works really well for our use case. Do you see any downside of getting rid of our rest endpoints though? does this mean the sdk is the only way to interact with the swap backend? not sure what the trpc experience is when used client side.

The old rest endpoint will remain for now but won't be updated to prevent the old SDKs from breaking. From there we can examine the use case for providing REST endpoints, but even so we'd have to figure out how to properly return the bigints that we now rely on `superjson` to return

PR #269, opened 2023-10-19
feat(sdk): open deposit channel from SDK

acdibble:


packages/shared/src/broker.ts
```diff
@@ -132,53 +106,32 @@ const makeRpcRequest = async <
...
-  }
-
-  return responseValidators[method].parse(body.result);
+  return responseValidators[method].parse(res.data.result);
 };
 
 export async function requestSwapDepositAddress(
   swapRequest: NewSwapRequest,
-  opts?: { url: string; commissionBps: number },
+  opts: { url: string; commissionBps: number },
```
jerryafr, line 121: These changes seem not to be backward compatible, is it ok to do them?

jerryafr (approved):

packages/shared/src/broker.ts
acdibble: Yeah, the `shared` package is an internal library. These functions aren't exposed directly to the outside world.

PR #276, opened 2023-10-23
feat(swap): [WEB-371] calculate and expose depositChannelExpiryTime

jerryafr:
[WEB-371]

- calculate and expose depositChannelExpiryTime


packages/swap/src/routes/swap.ts
```diff
@@ -132,6 +133,20 @@ router.get(
       );
     }
 
+    let depositChannelExpiryTime = new Date();
```
acdibble, line 136:
```suggestion
    let depositChannelExpiryTime: Date | null = null;
```

packages/swap/src/routes/swap.ts
```diff
@@ -132,6 +133,20 @@ router.get(
...
+    if (swapDepositChannel && !swapDepositChannel.isExpired) {
```
acdibble, line 137: if it expired, I'd expect to receive the time it expired at or nothing I think

packages/swap/src/routes/swap.ts
```diff
@@ -182,10 +198,26 @@ router.post(
...
+      ]);
+
+    const depositChannelExpiryTime = calculateExpiryTime({
+      chain: result.data.srcChain,
+      startBlock: chainInfo?.height,
+      expiryBlock: sourceChainExpiryBlock,
+    });
+
+    res.json({
+      ...response,
```
acdibble, line 218: don't want to lose this
```suggestion
      sourceChainExpiryBlock,
      ...response,
```

packages/swap/src/utils/function.ts
```diff
@@ -31,3 +32,35 @@ export const handleExit = (cb: AnyFunction) => {
...
+      break;
+    case 'Ethereum':
+      diff *= 15;
+      break;
+    case 'Polkadot':
+      diff *= 6;
+      break;
+    default:
+      unreachable(chain, 'Unsupported chain');
+  }
```
acdibble, line 64: would be terser to put this into (yet another) map and look it up per chain

packages/swap/src/utils/function.ts
```diff
@@ -31,3 +32,35 @@ export const handleExit = (cb: AnyFunction) => {
...
+  startBlock?: bigint;
+  expiryBlock?: bigint | null;
+}) {
+  const { chain, startBlock, expiryBlock } = args;
+
+  if (startBlock == null || expiryBlock == null) {
+    return new Date();
+  }
+
+  let diff = Number(expiryBlock - startBlock);
```
acdibble, line 51: a little more descriptive:
```suggestion
  let blocksRemaining = Number(expiryBlock - startBlock);
```

packages/swap/src/routes/swap.ts
jerryafr: As the time long in the past may be too inaccurate, will send null for expired ones

packages/swap/src/routes/swap.ts
```diff
@@ -132,6 +133,20 @@ router.get(
...
+    let depositChannelExpiryTime: Date | null = null;
+    if (swapDepositChannel && !swapDepositChannel.isExpired) {
+      const chainInfo = await prisma.chainTracking.findFirst({
+        where: {
+          chain: swapDepositChannel?.srcChain,
+        },
+      });
+      depositChannelExpiryTime = calculateExpiryTime({
+        chain: swapDepositChannel.srcChain,
+        startBlock: chainInfo?.height,
```
GabrielBuragev, line 145: can we not simply use the `swapDepositChannel.issuedBlock` instead of `chainInfo.height` and always return a timestamp with the expected expiration time?

jerryafr: issued block is not the issued block in source chain. Expiry on the other hand happens based on the source chain's block

GabrielBuragev: Ah, now it makes sense.

packages/swap/prisma/schema.prisma
```diff
@@ -194,3 +194,12 @@ model Pool {
   @@unique([baseAsset, pairAsset])
   @@schema("public")
 }
+
+model ChainTracking {
+  id        Int      @id @default(autoincrement())
+  chain     Chain    @unique
+  height    BigInt   @default(0)
+  updatedAt DateTime @updatedAt @db.Timestamptz(3)
```
GabrielBuragev, line 202: anyone knows what the `3` in `Timestampz(3)` means?

GabrielBuragev (approved):

packages/swap/prisma/schema.prisma
GabrielBuragev: Got it:
> time, timestamp, and interval accept an optional precision value p which specifies the number of fractional digits retained in the seconds field. By default, there is no explicit bound on precision. The allowed range of p is from 0 to 6.

packages/swap/src/utils/function.ts
```diff
@@ -31,3 +32,32 @@ export const handleExit = (cb: AnyFunction) => {
     exitHandlers.delete(cb);
   };
 };
+
+export function unreachable(value: never, message: string): never {
+  throw new Error(message);
+}
+
+const blockTimeMap: Record<Chain, number> = {
+  Bitcoin: 60,
```
zoheb391, line 41: bitcoin block time is ~10 min ?

packages/swap/src/routes/swap.ts
```diff
@@ -164,6 +179,7 @@ router.get(
       broadcastSucceededAt: swap?.egress?.broadcast?.succeededAt?.valueOf(),
       broadcastSucceededBlockIndex:
         swap?.egress?.broadcast?.succeededBlockIndex,
+      depositChannelExpiryTime: depositChannelExpiryTime?.valueOf(),
```
zoheb391, line 182: i think we should return the block as well. That allows us to show the expiry block and give the user an timed estimate for full transparency

zoheb391 (approved):

packages/swap/src/utils/function.ts
```diff
@@ -31,3 +32,32 @@ export const handleExit = (cb: AnyFunction) => {
     exitHandlers.delete(cb);
   };
 };
+
+export function unreachable(value: never, message: string): never {
```
zoheb391, line 36: unused?

packages/swap/src/routes/swap.ts
jerryafr: It's not in the design for now, but guys are updating the design. If needed, I will add it too

PR #280, opened 2023-10-27
fix(swap): [WEB-371] save srcChainExpiryBlock on opening deposit channel

jerryafr:
[WEB-371]
- save srcChainExpiryBlock on opening deposit channel

### DIFF
--- a/packages/swap/src/event-handlers/swapDepositAddressReady.ts
+++ b/packages/swap/src/event-handlers/swapDepositAddressReady.ts
@@ -35,6 +35,19 @@ export const swapDepositAddressReady = async ({
     ...rest
   } = swapDepositAddressReadyArgs.parse(event.args);
 
+  const data = {
+    srcChain: depositAddress.chain,
+    srcAsset: sourceAsset,
+    depositAddress: depositAddress.address,
+    expectedDepositAmount: 0,
+    destAsset: destinationAsset,
+    destAddress: destinationAddress.address,
+    srcChainExpiryBlock: sourceChainExpiryBlock,
+    issuedBlock,
+    channelId,
+    ...rest,
+  };
+
   await prisma.swapDepositChannel.upsert({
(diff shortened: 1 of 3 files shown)


packages/swap/src/event-handlers/swapDepositAddressReady.ts
```diff
@@ -43,19 +56,8 @@ export const swapDepositAddressReady = async ({
...
-      destAsset: destinationAsset,
-      destAddress: destinationAddress.address,
-      srcChainExpiryBlock: sourceChainExpiryBlock,
-      issuedBlock,
-      channelId,
-      ...rest,
-    },
-    update: {},
+    create: data,
+    update: data,
```
GabrielBuragev, line 60: Do we need to update the entry if it exists here? I think we are trying to avoid doing this.

jerryafr: There was some data missing and since we were not updating at all, the missing data was never filled.
Since what we get from the state chain is the source of truth, I think it makes sense to update the data.
From performance perspective, I don't think it has so much negativity

acdibble (approved):

packages/swap/src/handlers/openSwapDepositChannel.ts
```diff
@@ -51,6 +51,7 @@ export default async function openSwapDepositChannel(
       create: {
         ...rest,
         depositAddress,
+        srcChainExpiryBlock: sourceChainExpiryBlock,
```
niklasnatter, line 54: would have expected this to trigger a test, but i guess we dont test the api? 🤔

jerryafr: this was not tested by the api tests, however I've added a simple unit test for it now

niklasnatter (approved):

PR #287, opened 2023-10-30
fix(swap): [WEB-371] deposit amount is overwritten with 0 in swapDepositAddressReady

jerryafr:
[WEB-371]
- deposit amount is overwritten with zero in swapDepositAddressReady


niklasnatter (approved):

packages/swap/src/event-handlers/swapDepositAddressReady.ts
```diff
@@ -56,7 +55,10 @@ export const swapDepositAddressReady = async ({
         srcChain: depositAddress.chain,
       },
     },
-    create: data,
+    create: {
+      expectedDepositAmount: 0,
```
niklasnatter, line 59: maybe add a test to make sure we dont do this again 🙂

acdibble (approved):

PR #302, opened 2023-11-02
feat: use new RPCs [WEB-474][WEB-402]

acdibble:
Fixes WEB-474
Fixes WEB-402


packages/swap/src/utils/rpc.ts
```diff
@@ -0,0 +1,47 @@
...
+    BTC: poolsEnv.fees.Bitcoin.BTC.poolInfo,
+    ETH: poolsEnv.fees.Ethereum.ETH.poolInfo,
+    USDC: poolsEnv.fees.Ethereum.USDC.poolInfo,
+    FLIP: poolsEnv.fees.Ethereum.FLIP.poolInfo,
+    DOT: poolsEnv.fees.Polkadot.DOT.poolInfo,
+  };
+  return fees;
+};
+
+export type PoolFees = Awaited<
```
zoheb391, line 45: unused?

packages/shared/src/rpc/index.ts
```diff
@@ -0,0 +1,168 @@
...
+    method: string,
+    responseParser: R,
+    paramsParser?: P,
+  ) =>
+  async (
+    network: ChainflipNetwork,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    params: z.input<P> extends any ? void : z.input<P>,
+  ): Promise<CamelCaseRecord<z.output<R>>> => {
+    const { data } = await axios.post(`https://${network}.chainflip.xyz`, {
```
zoheb391, line 44: is this always `.xyz` ?

packages/shared/src/broker.ts
```diff
@@ -124,6 +124,8 @@ const makeRpcRequest = async <
     params: requestValidators[method].parse(params),
   });
 
+  console.log(res);
```
zoheb391, line 127: missed

packages/sdk/src/swap/sdk.ts
```diff
@@ -43,6 +44,8 @@ export class SwapSDK {
 
   private readonly brokerConfig?;
 
+  private environment?: Environment;
```
zoheb391, line 47: not sure if i like the name environment here. I think it can be something more unique to this situation. i.e `stateChainEnvironment` or  for something shorter `constants`

niklasnatter: i like stateChainEnvironment

packages/shared/src/consts.ts
```diff
@@ -8,38 +8,6 @@ const POOLS_NETWORK_FEE_HUNDREDTH_PIPS: Partial<
 export const getPoolsNetworkFeeHundredthPips = (network: ChainflipNetwork) =>
   POOLS_NETWORK_FEE_HUNDREDTH_PIPS[network] ?? 0;
 
-// TODO: fetch minimum deposit amounts via rpc from the state chain
```
niklasnatter, line 11: 🥳

packages/shared/src/consts.ts
```diff
@@ -8,38 +8,6 @@ const POOLS_NETWORK_FEE_HUNDREDTH_PIPS: Partial<
 export const getPoolsNetworkFeeHundredthPips = (network: ChainflipNetwork) =>
```
niklasnatter, line 8: think we can replace this with the implementation in `packages/swap/src/utils/rpc.ts`?

packages/swap/src/utils/rpc.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const getMinimumSwapAmount = async (
+  network: ChainflipNetwork,
+  asset: Asset,
+): Promise<bigint> => {
+  const swapEnv = await getSwappingEnvironment(network);
+  return readAssetValue(swapEnv.minimumSwapAmounts, asset);
+};
+
+export const getPoolsNetworkFeeHundredthPips = async (
```
niklasnatter, line 31: not sure about the name - it looks like it is returning pool infos, not the network fee (which should be the same for all pools)?

niklasnatter: also i dont think its used atm right?

package.json
```diff
@@ -10,7 +10,6 @@
     "test:ci:shared": "pnpm --filter sdk-shared test:ci",
     "test:ci:sdk": "pnpm --filter sdk test:ci",
     "test:ci:swap": "pnpm --filter swap test:ci",
-    "test:ci:cli": "pnpm --filter cli test:ci",
```
niklasnatter, line 13: 🧐

packages/sdk/src/swap/assets.ts
```diff
@@ -8,63 +7,74 @@ import {
...
   contractAddress: undefined,
   decimals: assetDecimals[Assets.ETH],
   name: 'Ether',
   symbol: 'ETH',
   isMainnet: !isTestnet(network),
-  minimumDepositAmount: getMinimumDepositAmount(network, Assets.ETH),
-  minimumSwapAmount: getMinimumSwapAmount(network, Assets.ETH),
+  minimumDepositAmount:
+    env.ingressEgress.minimumDepositAmounts.Ethereum.ETH.toString(),
+  minimumSwapAmount: env.swapping.minimumSwapAmounts.Ethereum.ETH.toString(),
```
acdibble, line 27: I doubt we need both of these because the `minimumSwapAmount` should always be greater than the deposit amount

niklasnatter: i think the reason for the separate `minimumDepositAmount` is that its more expensive for the protocol to "fetch" the funds than just swap them. but could remember this wrong

packages/shared/src/rpc/index.ts
```diff
@@ -0,0 +1,181 @@
...
+  ): Promise<CamelCaseRecord<z.output<R>>> => {
+    const url =
+      'network' in urlOrNetwork
+        ? RPC_URLS[urlOrNetwork.network]
+        : urlOrNetwork.rpcUrl;
+    const { data } = await axios.post(url, {
+      jsonrpc: '2.0',
+      method,
+      params: paramsParser?.parse(params),
+      id: 1,
```
niklasnatter, line 61: think this should change between requests

packages/shared/src/rpc/index.ts
```diff
@@ -0,0 +1,181 @@
...
+  redemption_tax: numberOrHex,
+  minimum_funding_amount: numberOrHex,
+});
+export const getFundingEnvironment = createRequest(
+  'cf_funding_environment',
+  fundingEnvironment,
+);
+
+const chainAssetMap = <Z extends z.ZodTypeAny>(parser: Z) =>
+  z.object({
```
niklasnatter, line 83: do we need to hardcode these object shapes or is there a way we could generate this (and the `rpcAsset`/`feesInfo`) from the constants we have?

packages/shared/src/schemas.ts
```diff
@@ -25,15 +25,30 @@ export const ccmMetadataSchema = z.object({
...
-  })
+  .intersection(
+    z.object({
+      destAddress: z.string(),
+      amount: numericString,
+      ccmMetadata: ccmMetadataSchema.optional(),
+    }),
+    z.union([
+      z
+        .object({
```
niklasnatter, line 36: this is the old way right? should we add a comment that we can remove it in the future?

packages/shared/src/rpc/index.ts
niklasnatter: its okay for me to merge it like this

packages/cli/package.json
```diff
@@ -26,8 +26,6 @@
   ],
   "scripts": {
     "build": "rm -rf dist && tsup",
-    "test": "jest",
```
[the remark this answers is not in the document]
acdibble: They were broken and they don't give us much value imo. Can look into restoring them if you think it's worth it but I'm doubtful

packages/shared/src/rpc/index.ts
acdibble: The ID only matters for websocket connections. For an HTTP request, there is only ever 1 request so there is no need to change it

packages/sdk/src/swap/assets.ts
acdibble: I think for LP deposit channels the minimum deposit amount is more interesting, but here it should be redundant.

packages/shared/src/rpc/index.ts
acdibble: I thought about trying to generate it but I don't think we will be adding assets fast enough to warrant the time it would take to try to get the zod parser to output exactly what we want

packages/shared/src/rpc/index.ts
```diff
@@ -0,0 +1,181 @@
...
+  minimum_deposit_amounts: chainAssetNumberMap,
+});
+
+export const getIngressEgressEnvironment = createRequest(
+  'cf_ingress_egress_environment',
+  ingressEgressEnvironment,
+);
+
+const rpcAsset = z.union([
+  z.literal('Btc'),
```
niklasnatter, line 138: this should not be returned from the rpc anymore, right?

packages/shared/src/rpc/index.ts
```diff
@@ -0,0 +1,181 @@
...
+      jsonrpc: '2.0',
+      method,
+      params: paramsParser?.parse(params),
+      id: 1,
+    });
+
+    const result = responseParser.safeParse(data.result);
+
+    if (result.success) {
+      return camelCaseKeys(result.data);
```
niklasnatter, line 67: i always found it a bit weird that return values are camel cased automatically, so i would be okay with skipping this. but no strong opinion

jest.config.mjs
```diff
@@ -123,7 +123,7 @@ export default {
   // reporters: undefined,
 
   // Automatically reset mock state before every test
-  // resetMocks: false,
+  resetMocks: true,
```
niklasnatter, line 126: is there a specific reason to change the default?

acdibble: Yeah there was cross contamination between the mocks. I would have turned this on from the beginning if I knew it wasn't

packages/shared/src/rpc/index.ts
acdibble: wtf happened

niklasnatter (approved):

PR #303, opened 2023-11-02
chore: Make clear that expiry time is estimated and add expiry block to swap status

niklasnatter:
Relevant changes in `packages/sdk/src/swap/types.ts`: https://github.com/chainflip-io/chainflip-sdk-monorepo/pull/303/files#diff-06c5a42eb9e83f26d636e2de6a76033453c31133e126100a3e7bf259f2e19ad9


packages/sdk/src/swap/types.ts
```diff
@@ -57,7 +57,8 @@ export interface CommonStatusFields extends ChainsAndAssets {
   depositAddress: string | undefined;
   depositChannelCreatedAt: number | undefined;
   expectedDepositAmount: string | undefined;
-  depositChannelExpiryTime: number | undefined;
+  depositChannelExpiryBlock: bigint;
```
niklasnatter, line 60: this info might be useful for consumers that dont want to rely on the time estimation. also we already expose it from the `requestDepositAddress` method

packages/swap/src/routes/swap.ts
acdibble, line 1: this route handler is deprecated in favor of the TRPC handler, you don't have to update this anymore

niklasnatter: most of the changes are in the status api (`get` handler) - that one is still used, right?

niklasnatter: changed the property name in the `post` handler only to make typescript happy

zoheb391 (approved):

acdibble (approved):

PR #308, opened 2023-11-03
fix(swap,sdk): return estimated expiry time

jerryafr:


zoheb391 (approved):

packages/sdk/src/swap/types.ts
```diff
@@ -59,6 +59,7 @@ export interface CommonStatusFields extends ChainsAndAssets {
   expectedDepositAmount: string | undefined;
   depositChannelExpiryBlock: bigint;
   estimatedDepositChannelExpiryTime: number | undefined;
+  isExpired: boolean;
```
niklasnatter, line 62: as this is the status response, i think we need to be explicit that this is about the deposit channel
```suggestion
  depositChanneIsExpired: boolean;
```

packages/swap/src/routes/swap.ts
```diff
@@ -183,6 +183,7 @@ router.get(
         swapDepositChannel?.srcChainExpiryBlock?.toString(),
       estimatedDepositChannelExpiryTime:
         estimatedDepositChannelExpiryTime?.valueOf(),
+      isExpired: swapDepositChannel?.isExpired ?? false,
```
niklasnatter, line 186: i would say its fine to have this as `undefined` if there is no deposit channel for the swap
```suggestion
      isExpired: swapDepositChannel?.isExpired,
```

niklasnatter (approved):

PR #315, opened 2023-11-07
feat(sdk, swap): add trpc for getStatus [WEB-289]

zoheb391:


packages/swap/src/routes/swap.ts
acdibble, line 1: can't remove these yet because they will break the existing SDK

acdibble (approved):

packages/swap/src/handlers/getSwapStatus.ts
```diff
@@ -0,0 +1,191 @@
...
+    state = swapState.enum.SWAP_EXECUTED;
+  } else if (swap?.depositReceivedAt) {
+    state = swapState.enum.DEPOSIT_RECEIVED;
+  } else {
+    state = swapState.enum.AWAITING_DEPOSIT;
+  }
+
+  const srcAsset = readField(swap, swapDepositChannel, 'srcAsset');
+  const destAsset = readField(swap, swapDepositChannel, 'destAsset');
+  assert(srcAsset && destAsset, 'srcAsset and destAsset should be defined');
```
zoheb391, line 120: i added this assert because the type was `Asset | undefined` But in practice I don't think it can be undefined if we have a deposit channel or swap

packages/swap/src/handlers/getSwapStatus.ts
```diff
@@ -0,0 +1,191 @@
+import assert from 'assert';
```
zoheb391, line 1: for the most part this a copy of `/routes/swap.ts`

packages/shared/src/schemas.ts
```diff
@@ -39,6 +42,104 @@ export const openSwapDepositChannelSchema = z
...
+    .object({
+      gasBudget: string,
+      message: string.nullable(),
+    })
+    .optional()
+    .nullable(),
+});
+
+const awaitingState = baseSwapResponseSchema.extend({
+  state: swapState.refine((state) => state === swapState.Enum.AWAITING_DEPOSIT),
```
zoheb391, line 88: if anyone has a better idea how to solve this. pls let me know. I can't use a zod literal here because trpc was complaining. It isn't ideal right now because we still need to do `if(x in prop)`

packages/swap/src/handlers/getSwapStatus.ts
niklasnatter: can we reuse this in `/routes/swap.ts` like we did with the `openSwapDepositChannel`

packages/sdk/src/swap/types.ts
```diff
@@ -53,11 +53,15 @@ export interface SwapStatusRequest {
 }
 
 export interface CommonStatusFields extends ChainsAndAssets {
+  srcChain: Chain;
+  destChain: Chain;
+  srcAsset: Asset;
+  destAsset: Asset;
```
niklasnatter, line 59: this type extends `ChainsAndAssets` which should already contain this properties

packages/sdk/src/swap/types.ts
```diff
@@ -69,6 +73,7 @@ export interface CommonStatusFields extends ChainsAndAssets {
     | undefined;
 }
 
+// this is not 100% the ACTUAL response type. Its a visual guide as to what you can expect. The real type is inferred from the trpc output schema
```
niklasnatter, line 76: i would be okay with just ditching it

packages/shared/src/schemas.ts
niklasnatter: what was the trpc error message?

niklasnatter (commented):
nice! didnt know that trpc integrates zod directly

packages/swap/src/handlers/getSwapStatus.ts
zoheb391: we could but its not so simple because we'd need to serialize bigints. prefer to just switch over when we do

packages/sdk/src/swap/types.ts
zoheb391: i think we probably want it to display in docs.. unless we show zod schema in the docs

packages/shared/src/schemas.ts
zoheb391: complaining about the state return types not matching 🤷

packages/sdk/src/swap/types.ts
niklasnatter: we generate the type for the docs from the zod schema in other places, so we can do it here too 🙂 its in the `printTypes.mjs` script

packages/swap/src/handlers/getSwapStatus.ts
```diff
@@ -0,0 +1,191 @@
...
+      where: {
+        chain: swapDepositChannel.srcChain,
+      },
+    });
+    estimatedDepositChannelExpiryTime = calculateExpiryTime({
+      chain: swapDepositChannel.srcChain,
+      startBlock: chainInfo?.height,
+      expiryBlock: swapDepositChannel.srcChainExpiryBlock,
+    });
+  }
```
jerryafr, line 146: This is changed recently in the original `routes/swap.ts` file, can you please update it here too?

PR #321, opened 2023-11-09
fix(swap): change RPC urls

acdibble:


packages/swap/src/utils/rpc.ts
```diff
@@ -18,20 +13,11 @@ const readAssetValue = (
   return chainMinimums[asset.asset as keyof typeof chainMinimums];
 };
 
-export const getMinimumDepositAmount = async (
```
niklasnatter, line 21: not used anymore? 🤔

jerryafr (approved):

packages/swap/src/utils/rpc.ts
acdibble: was never used after a refactor in the original PR

PR #331, opened 2023-11-13
feat(sdk,swap): [WEB-499] set max byte size limit for ccmMessage

jerryafr:


packages/shared/src/parsers.ts
```diff
@@ -20,6 +20,13 @@ const errorMap: ZodErrorMap = (_issue, context) => ({
   message: `received: ${safeStringify(context.data)}`,
 });
 
+export const hexStringWithMaxByteSize = (maxByteSize: number) =>
+  z
+    .string()
+    .refine((val) => /^0x[0-9a-f]+$/.test(val))
```
acdibble, line 26: if possible:
```suggestion
  hexString
```

acdibble (approved):

packages/shared/src/__tests__/parsers.test.ts
```diff
@@ -87,3 +93,27 @@ describe('unsignedInteger', () => {
...
+  });
+
+  it('should reject non hex strings', () => {
+    expect(() => hexStringWithMaxByteSize(100).parse('hello')).toThrow(
+      'Invalid input',
+    );
+  });
+
+  it('should reject strings larger than max byte size', () => {
+    expect(() => hexStringWithMaxByteSize(3).parse('0xabcdef')).toThrow(
```
niklasnatter, line 115: i think i would expect this to pass as the value has 3 bytes?

packages/shared/src/parsers.ts
GabrielBuragev, line 26: Do we need to care about case-sensitivity here? 
I know that in the ethereum ecosystem, the wallet addresses are many times represented as `/^0x[0-9a-fA-F]+$/`. Users use all sorts of tools to copy these addresses and sometimes the result is, they have a checksummed address in their clipboard.
I guess it depends on what we expect to be passed as a value for the ccm but just wanted to raise the awareness on this one.

niklasnatter: good point! im actually not sure how the backend handles this - worth a check if it accepts uppercase letters

niklasnatter: yep sounds good - can we reuse/adjust the existing `hexString`?

jerryafr: updated it to reuse `hexString` which actually is case insensitive

niklasnatter (approved):

PR #380, opened 2023-12-07
feat: handle maximum swap amounts

acdibble:


packages/shared/src/rpc/__tests__/utils.test.ts
```diff
@@ -0,0 +1,60 @@
...
+const swappingEnv = {
+  minimumSwapAmounts: {
+    Ethereum: { ETH: 100000000000000000n, FLIP: 0n },
+  },
+  maximumSwapAmounts: {
+    Ethereum: { ETH: 1000000000000000000n, FLIP: null },
+  },
+} as any;
+
+describe(validateSwapAmount, () => {
```
niklasnatter, line 12: it would be great to add one integration test for the `getQuote` endpoint to make sure we dont remove that check by mistake and give quotes that will be wrong

zoheb391 (approved):

packages/shared/src/rpc/__tests__/utils.test.ts
acdibble: Added as well

PR #395, opened 2023-12-12
feat: Add fee information to swap status [WEB-482]

niklasnatter:


packages/shared/src/consts.ts
```diff
@@ -3,7 +3,10 @@ import { ChainflipNetwork, ChainflipNetworks } from './enums';
 const POOLS_NETWORK_FEE_HUNDREDTH_PIPS: Partial<
   Record<ChainflipNetwork, number>
 > = {
-  [ChainflipNetworks.perseverance]: 100,
+  [ChainflipNetworks.backspin]: 1000,
```
niklasnatter, line 6: unfortunately this is not available on any rpc yet

packages/swap/src/routes/__tests__/fee.test.ts
niklasnatter, line 1: this was never used. as we are reading the fee info from the rpc directly now we can remove it

packages/swap/src/event-handlers/swapExecuted.ts
```diff
@@ -15,16 +16,34 @@ export default async function swapExecuted({
...
     event.args,
   );
+  const swap = await prisma.swap.findUnique({
+    where: { nativeId: swapId },
+  });
+
+  // skip update if we are not tracking swap
+  if (!swap) {
+    return;
+  }
```
acdibble, line 29: we should be tracking all swaps now

niklasnatter: i think so, right? i was scared about breaking something but will remove the check

packages/swap/src/routes/swap.ts
```diff
@@ -156,6 +165,7 @@ router.get(
       egressAmount: swap?.egress?.amount?.toString(),
       egressScheduledAt: swap?.egress?.scheduledAt?.valueOf(),
       egressScheduledBlockIndex: swap?.egress?.scheduledBlockIndex,
+      paidFess: swap?.fees,
```
niklasnatter, line 168: add test and type

acdibble (approved):

packages/shared/src/consts.ts
niklasnatter: https://linear.app/chainflip/issue/WEB-683/use-network-fee-from-rpc-instead-of-hardcoding-it

packages/sdk/src/swap/types.ts
```diff
@@ -108,6 +108,7 @@ export type SwapStatusResponse = CommonStatusFields &
         egressAmount: string;
         egressScheduledAt: number;
         egressScheduledBlockIndex: string;
+        paidFess: SwapFee[];
```
acdibble, line 111:
```suggestion
        feesPaid: SwapFee[];
```

PR #398, opened 2023-12-13
feat: Add ingress and egress fees to quote response

niklasnatter:


jerryafr (approved):

packages/shared/src/rpc/index.ts
```diff
@@ -115,6 +115,8 @@ export const getSwappingEnvironment = createRequest(
 
 const ingressEgressEnvironment = z.object({
   minimum_deposit_amounts: chainAssetNumberMap,
+  ingress_fees: chainAssetNumberMap.optional(),
+  egress_fees: chainAssetNumberMap.optional(),
```
acdibble, line 119: We simply won't deploy this to an env w/o the new fields

```suggestion
  ingress_fees: chainAssetNumberMap,
  egress_fees: chainAssetNumberMap,
```

acdibble (approved):

PR #400, opened 2023-12-13
refactor: use min deposit if higher

acdibble:


packages/swap/src/__tests__/integration.test.ts
```diff
@@ -18,20 +18,28 @@ import { QuoteQueryParams } from '@/shared/schemas';
...
-
-    if (data.method === 'cf_ingress_egress_environment')
-      return Promise.resolve({ data: ingressEgressEnvironment('0x123') });
+    if (data.method === 'cf_environment') {
+      return Promise.resolve({
+        data: {
+          result: {
+            ...environment().result,
+            swapping: swappingEnvironment().result,
+            ingress_egress: ingressEgressEnvironment('0x123').result,
```
niklasnatter, line 38: no need for these values, we can just use `environment()`

packages/swap/src/routes/__tests__/quote.test.ts
```diff
@@ -37,15 +39,36 @@ jest.mock('@/shared/consts', () => ({
...
-    if (data.method === 'cf_swapping_environment')
-      return Promise.resolve({ data: swappingEnvironment() });
-
-    if (data.method === 'cf_ingress_egress_environment')
-      return Promise.resolve({ data: ingressEgressEnvironment('0x777') });
+    if (data.method === 'cf_environment') {
+      return Promise.resolve({
+        data: {
+          result: {
+            ...environment().result,
```
niklasnatter, line 46: same here but also okay to keep it i guess

packages/swap/src/routes/common.ts
```diff
@@ -34,6 +34,7 @@ export const asyncHandler = <
     try {
       await handler(req, res, next);
     } catch (error) {
+      console.error(error);
```
niklasnatter, line 37: 🙌

packages/shared/src/rpc/utils.ts
```diff
@@ -10,23 +11,38 @@ export const readAssetValue = <T>(
   return chainMinimums[asset.asset as keyof typeof chainMinimums];
 };
 
+export const readMinimumSwapAmount = (
```
niklasnatter, line 14: `getEffectiveMinimumSwapAmount`?

niklasnatter (approved):

packages/swap/src/__tests__/integration.test.ts
acdibble: yeah i was just trying to make it work fast. environment just clones the structure without injecting anything

packages/swap/src/routes/common.ts
niklasnatter: i actually liked the log lol

PR #401, opened 2023-12-14
feat(swap): add deposit channel table

acdibble:


niklasnatter (approved):

packages/swap/prisma/migrations/20231214103108_add_deposit_channel_table/migration.sql
```diff
@@ -0,0 +1,22 @@
+-- CreateTable
+CREATE TABLE "private"."DepositChannel" (
+    "id" BIGSERIAL NOT NULL,
+    "channelId" BIGINT NOT NULL,
+    "srcChain" "public"."Chain" NOT NULL,
+    "issuedBlock" INTEGER NOT NULL,
+    "depositAddress" TEXT NOT NULL,
+    "isSwapping" BOOLEAN NOT NULL,
```
jerryafr, line 8: would a type be better here maybe? this design only supports 2 types of channels

jerryafr (approved):

packages/swap/prisma/migrations/20231214103108_add_deposit_channel_table/migration.sql
acdibble: For the purpose of this database, we only care about if it is a swapping channel or not. If they add a third channel, we can break the boolean up if necessary

PR #406, opened 2023-12-15
chore(swap): Convert egress fee amount to egress asset in SwapEgressScheduled handler

niklasnatter:


acdibble (approved):

packages/swap/src/event-handlers/__tests__/swapEgressScheduled.test.ts
```diff
@@ -14,17 +15,29 @@ const {
...
-  ),
+  post: jest.fn((url, data) => {
+    if (data.method === 'cf_environment') {
+      return Promise.resolve({ data: environment({ egressFee: '0x777777' }) });
+    }
+
+    if (data.method === 'cf_swap_rate') {
+      return Promise.resolve({
+        data: swapRate({
+          output: `0x${(BigInt(data.params[2]) * 3n).toString(16)}`,
```
GabrielBuragev, line 26: why `* 3n`?

niklasnatter: just a random rate that could be the rate of the pool

GabrielBuragev (approved):

jerryafr (approved):

packages/swap/src/event-handlers/__tests__/swapEgressScheduled.test.ts
```diff
@@ -90,4 +103,71 @@ describe(swapEgressScheduled, () => {
       fees: [{ id: expect.any(BigInt), swapId: expect.any(BigInt) }],
     });
   });
+
+  it('updates an existing swap when the egress asset is not the native asset', async () => {
```
GabrielBuragev, line 107: Can we maybe check if the `getSwapRate` function was called in this test?

niklasnatter: the snapshot checks makes sure that the egress fee is three times the fee of the environment 🙂 but will add another assertion in the next pr

PR #409, opened 2023-12-15
fix(swap): Remove duplicated egress amount column

niklasnatter:


packages/swap/prisma/schema.prisma
```diff
@@ -90,7 +90,6 @@ model Swap {
   srcAmount                    Decimal             @db.Decimal(30, 0)
   intermediateAmount           Decimal?            @db.Decimal(30, 0)
   destAmount                   Decimal?            @db.Decimal(30, 0)
-  egressAmount                 Decimal?            @db.Decimal(30, 0)
```
niklasnatter, line 93: the amount is already stored on the `Egress` entity

packages/swap/src/event-handlers/swapEgressScheduled.ts
```diff
@@ -107,18 +107,25 @@ export default async function swapEgressScheduled({
   // TODO: use an accurate source for determining the egress fee once the protocol provides one
   const egressFee = await getEgressFeeAtBlock(block.hash, swap.destAsset);
 
-  await prisma.swap.update({
-    where: { nativeId: swapId },
-    data: {
-      egressAmount: swap.destAmount?.sub(egressFee.toString()),
```
niklasnatter, line 113: this is what happens if you have two columns for the same thing, you forget to update one

acdibble (approved):

PR #422, opened 2023-12-18
fix(swap): Use toFixed to stringify decimals.js number

niklasnatter:


packages/swap/src/routes/__tests__/swap.test.ts
```diff
@@ -98,7 +99,7 @@ describe('server', () => {
           "destAsset": "DOT",
           "destChain": "Polkadot",
           "estimatedDepositChannelExpiryTime": 1699527900000,
-          "expectedDepositAmount": "10000000000",
+          "expectedDepositAmount": "25000000000000000000000",
```
niklasnatter, line 102: this number would be stringified as `2.5e+22` without the change

acdibble (approved):

jerryafr (approved):

PR #423, opened 2023-12-18
fix(swap): Use bech32-buffer to decode/encode bitcoin addresses

niklasnatter:


packages/shared/src/bitcoin.ts
```diff
@@ -0,0 +1,35 @@
...
+      break;
+    case 'backspin':
+      hrp = 'bcrt';
+      break;
+    case 'berghain':
+    case 'mainnet':
+      hrp = 'bc';
+      break;
+    default:
+      hrp = 'tb';
```
acdibble, line 26:
```suggestion
      throw new Error('CHAINFLIP_NETWORK not set');
```

acdibble (approved):

GabrielBuragev (approved):

PR #425, opened 2023-12-20
chore(swap): Use CHAINFLIP_NETWORK env variable to determine appropriate address validation

niklasnatter:


packages/shared/src/validation/addressValidation.ts
```diff
@@ -108,25 +108,32 @@ export const validateBitcoinRegtestAddress: AddressValidator = (
...
-  isMainnet = true,
+  network: ChainflipNetwork,
 ): boolean => {
   switch (chain) {
     case 'Bitcoin':
-      return isMainnet
-        ? validateBitcoinMainnetAddress(address)
-        : validateBitcoinTestnetAddress(address) ||
-            validateBitcoinRegtestAddress(address);
+      switch (network) {
```
acdibble, line 115: looks like an opportunity for another map: `Record<ChainflipNetwork, Record<ChainflipChain, (addr: string) => boolean>>`

packages/swap/src/handlers/openSwapDepositChannel.ts
```diff
@@ -12,7 +12,13 @@ import ServiceError from '../utils/ServiceError';
 export default async function openSwapDepositChannel(
   input: z.output<typeof openSwapDepositChannelSchema>,
 ) {
-  if (!validateAddress(input.destChain, input.destAddress, isProduction)) {
+  if (
+    !validateAddress(
+      input.destChain,
+      input.destAddress,
+      process.env.CHAINFLIP_NETWORK as ChainflipNetwork,
```
acdibble, line 19: could do an zod env migration here

acdibble (approved):
was also thinking about this last night

packages/swap/src/handlers/openSwapDepositChannel.ts
niklasnatter: yes this should be the next app we tackle 🙂

GabrielBuragev (approved):

packages/shared/src/validation/addressValidation.ts
niklasnatter: it looks like that indeed, but im too lazey to do right now 😇

PR #450, opened 2024-01-10
fix(service): Ignore deposit if there is no swap for a swap deposit channel

niklasnatter:


jerryafr (approved):

packages/swap/src/event-handlers/networkDepositReceived.ts
```diff
@@ -71,6 +71,17 @@ export const networkDepositReceived =
...
+      logger.warn('no swap found for deposit to swap deposit channel', {
+        block: block.height,
+        eventIndexInBlock: event.indexInBlock,
+        eventName: event.name,
+        depositAddress,
+      });
+
+      return;
+    }
     assert(swap, 'swap not found for deposit');
```
GabrielBuragev, line 85: do we still need this?

GabrielBuragev (approved):

packages/swap/src/event-handlers/networkDepositReceived.ts
niklasnatter: no, good catch!

PR #457, opened 2024-01-16
chore(sdk): Make minimum_swap_amounts property of swapping environment optional

niklasnatter:
Preparation for https://github.com/chainflip-io/chainflip-backend/pull/4419

### DIFF
--- a/packages/sdk/src/swap/assets.ts
+++ b/packages/sdk/src/swap/assets.ts
@@ -7,7 +7,6 @@ import {
   isTestnet,
 } from '@/shared/enums';
 import type { Environment } from '@/shared/rpc';
-import { readMinimumSwapAmount } from '@/shared/rpc/utils';
 import type { AssetData } from './types';
 
 type AssetFn = (
@@ -24,10 +23,8 @@ export const eth$: AssetFn = (network, env) => ({
   symbol: 'ETH',
   chainflipId: 'ETH',
   isMainnet: !isTestnet(network),
-  minimumSwapAmount: readMinimumSwapAmount(env, {
-    chain: 'Ethereum',
-    asset: 'ETH',
-  }).toString(),
+  minimumSwapAmount:
+    env.ingressEgress.minimumDepositAmounts.Ethereum.ETH.toString(),
   maximumSwapAmount:
     env.swapping.maximumSwapAmounts.Ethereum.ETH?.toString() ?? null,
 });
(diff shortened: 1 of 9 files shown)


acdibble (commented):
we can just remove it. we also have to publish an SDK asap after making this change

niklasnatter:
> we can just remove it. we also have to publish an SDK asap after making this change

You are right, removing it is probably better. Will need to touch a few files

packages/sdk/src/swap/assets.ts
```diff
@@ -24,10 +23,8 @@ export const eth$: AssetFn = (network, env) => ({
   symbol: 'ETH',
   chainflipId: 'ETH',
   isMainnet: !isTestnet(network),
-  minimumSwapAmount: readMinimumSwapAmount(env, {
-    chain: 'Ethereum',
-    asset: 'ETH',
-  }).toString(),
+  minimumSwapAmount:
```
niklasnatter, line 26: should we rename this property of the `TokenData` to `minimumDepositAmount`?

packages/sdk/src/swap/sdk.ts
```diff
@@ -201,15 +201,14 @@ export class SwapSDK {
...
     minimumDepositAmounts: ChainAssetMap<bigint>;
   }> {
     const {
-      swapping: { minimumSwapAmounts, maximumSwapAmounts },
+      swapping: { maximumSwapAmounts },
       ingressEgress: { minimumDepositAmounts },
     } = await this.getStateChainEnvironment();
 
-    return { minimumSwapAmounts, maximumSwapAmounts, minimumDepositAmounts };
+    return { maximumSwapAmounts, minimumDepositAmounts };
```
niklasnatter, line 212: this is a small bc break, how should we handle it?

packages/sdk/src/swap/assets.ts
acdibble: No, for this context it is the minimum swap amount

packages/sdk/src/swap/sdk.ts
acdibble, line 212:
```suggestion
    return { maximumSwapAmounts, minimumSwapAmounts: minimumDepositAmounts };
```

acdibble: update docs and send some messages in the discord

jerryafr: this is just to retain backward compatibility?
If so, maybe breaking it is actually not a bad idea. It will force them to consider correct implementation

jerryafr (approved):

acdibble (approved):

packages/sdk/src/swap/sdk.ts
acdibble: If it were a general purpose SDK, I think we could keep the name, but for a swapping SDK, this name is more fitting. I don't have too strong an opinion, but it should probably match the name of the field on the tokens

niklasnatter: agree with this - changed

PR #460, opened 2024-01-16
feat(swap): read pending deposits from redis [WEB-717]

acdibble:


packages/swap/src/deposit-tracking/index.ts
```diff
@@ -1,19 +1,40 @@
...
+    prisma.chainTracking.findFirst({ where: { chain } }),
+  ]);
+
+  if (deposits.length === 0 || tracking == null) return undefined;
+
+  const confirmations =
+    tracking.height - BigInt(deposits[0].deposit_chain_block_height);
+
+  return {
+    amount: deposits[0].amount.toString(),
```
jerryafr, line 37: what if there are multiple deposits with the same address?

acdibble: It's an edge case that I'm not sure there's a good way of handling. For the purposes of our swapping app, this oldest deposit will be the one that will cause the app to progress forward in the flow.

We need to refactor our deposit channel status flow to return multiple swaps and stuff. I think at that point we could handle the case of multiple deposits but right now I can't think of much we can really do

packages/shared/src/node-apis/redis.ts
```diff
@@ -74,11 +74,15 @@ export default class RedisClient {
     return value ? broadcastParsers[chain].parse(JSON.parse(value)) : null;
   }
 
-  async getDeposits(chain: Chain, address: string) {
+  async getDeposits(chain: Chain, asset: Asset, address: string) {
     const parsedAddress = chain === 'Polkadot' ? ss58ToHex(address) : address;
     const key = `deposit:${chain}:${parsedAddress}`;
     const value = await this.client.get(key);
-    return value ? deposits.parse(JSON.parse(value)) : [];
+    return value
```
niklasnatter, line 81: is the order of this list guaranteed? might make sense to sort it by `deposit_chain_block_height`

niklasnatter (approved):

packages/shared/src/node-apis/redis.ts
acdibble: done

PR #461, opened 2024-01-16
fix(swap): Fix unknown property when storing deposit channel with ccm metadata to database

niklasnatter:


packages/swap/src/handlers/__tests__/openSwapDepositChannel.test.ts
```diff
@@ -17,34 +17,33 @@ jest.mock('../../utils/screenAddress', () => ({
...
+  it('creates channel and stores it in the database', async () => {
+    jest.mocked(axios.post).mockResolvedValueOnce({ data: environment() });
     jest.mocked(broker.requestSwapDepositAddress).mockResolvedValueOnce({
       sourceChainExpiryBlock: BigInt('1000'),
       address: 'address',
       channelId: BigInt('888'),
       issuedBlock: 123,
     });
 
-    prisma.swapDepositChannel.upsert = jest.fn().mockResolvedValueOnce({
```
niklasnatter, line 36: i changed this to use the database because this approach would not have caught this bug (because it does not invoke the db code)

packages/swap/src/handlers/openSwapDepositChannel.ts
```diff
@@ -39,7 +39,7 @@ export default async function openSwapDepositChannel(
     env.CHAINFLIP_NETWORK,
   );
 
-  const { destChain, ...rest } = input;
+  const { destChain, ccmMetadata, ...rest } = input;
```
niklasnatter, line 42: we passed the `ccmMetadata` to the `prisma.swapDepositChannel.upsert` call which resulted in an error because  `ccmMetadata` is not a column in the database

jerryafr (approved):

packages/swap/src/handlers/openSwapDepositChannel.ts
acdibble: We don't want to save it in the database?

niklasnatter: we want in the future, but the table does not have the columns right now. will save them in a follow up, but for now i would like to fix the working example in the docs 🙂

PR #468, opened 2024-01-18
refactor(swap): handle API changes, fix lrange

acdibble:


packages/shared/src/node-apis/redis.ts
```diff
@@ -46,13 +43,15 @@ type PolkadotBroadcast = ChainBroadcast<'Polkadot'>;
...
-  value: number.transform((value) =>
-    new BigNumber(value).shiftedBy(assetDecimals.BTC).toString(),
-  ),
-  tx_hash: string.transform((value) => `0x${value}` as const),
-});
+const mempoolTransaction = jsonString.pipe(
+  z.object({
+    confirmations: number,
+    value: number.transform((value) =>
+      new BigNumber(value).shiftedBy(assetDecimals.BTC).toString(),
```
niklasnatter, line 50: i think you changed this, do we still need the shift?

packages/shared/src/node-apis/redis.ts
```diff
@@ -89,19 +88,17 @@ export default class RedisClient {
...
-      ? deposits
-          .parse(JSON.parse(value))
-          .filter((deposit) => deposit.asset.asset === asset)
-          .sort(sortDepositAscending)
-      : [];
+    const deposits = await this.client.lrange(key, 0, -1);
+    return deposits
+      .map((deposit) => depositSchema.parse(deposit))
+      .filter((deposit) => deposit.asset === asset)
+      .sort(sortDepositAscending);
```
niklasnatter, line 95: as this is a linked list and we are using `rpush` to store it, i think we would not need to sort

acdibble: yeah we probably never needed to but we're not guaranteed to always be using `rpush` to store it

packages/shared/src/node-apis/redis.ts
acdibble: fixed

jerryafr (approved):

niklasnatter (approved):
cool!

PR #470, opened 2024-01-18
chore(swap): Make expectedDepositAmount nullable and add BROADCASTED status to exposed type

niklasnatter:


packages/swap/src/event-handlers/swapDepositAddressReady.ts
```diff
@@ -79,7 +79,6 @@ export const swapDepositAddressReady = async ({
         },
       },
       create: {
-        expectedDepositAmount: 0,
```
niklasnatter, line 82: i think its better to make it explicit that we dont know what the user intends to send. before this change, we exposed `"expectedDepositAmount": 0` from the status in the SDK

packages/sdk/src/swap/types.ts
```diff
@@ -125,6 +125,22 @@ export type SwapStatusResponse = CommonStatusFields &
...
+        intermediateAmount: string | undefined;
+        swapExecutedAt: number;
+        swapExecutedBlockIndex: string;
+        egressAmount: string;
+        egressScheduledAt: number;
+        egressScheduledBlockIndex: string;
+        feesPaid: SwapFee[];
+        broadcastRequestedAt: number;
+        broadcastRequestedBlockIndex: string;
+      }
```
acdibble, line 143: I think this can just be folded into the `BROADCAST_REQUESTED`, i.e.: `state: 'BROADCAST_REQUESTED' | 'BROADCASTED'`

acdibble (approved):

jerryafr (approved):

PR #474, opened 2024-01-19
fix(swap): handle numbers [WEB-717]

acdibble:
`NumberOrHex` can be serialized into a number if the underlying number is a `u64` and the value is < `Number.MAX_SAFE_INTEGER`

### DIFF
--- a/packages/shared/src/parsers.ts
+++ b/packages/shared/src/parsers.ts
@@ -70,7 +70,7 @@ export const ethereumAddress = hexString.refine((address) =>
 export const u64 = numericString.transform((arg) => BigInt(arg));
 
 export const u128 = z
-  .union([numericString, hexString])
+  .union([number, z.union([numericString, hexString])])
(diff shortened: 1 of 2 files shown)


niklasnatter (approved):

packages/shared/src/parsers.ts
```diff
@@ -70,7 +70,7 @@ export const ethereumAddress = hexString.refine((address) =>
 export const u64 = numericString.transform((arg) => BigInt(arg));
 
 export const u128 = z
-  .union([numericString, hexString])
+  .union([number, z.union([numericString, hexString])])
```
jerryafr, line 73: wouldn't this work? 🤔 
```suggestion
  .union([number, numericString, hexString])
```

jerryafr (approved):

packages/shared/src/parsers.ts
acdibble: true

PR #497, opened 2024-01-24
feat(swap, sdk): handle deposit ignored event

zoheb391:


packages/swap/src/event-handlers/networkDepositReceived.ts
```diff
@@ -12,7 +16,7 @@ import { EventHandlerArgs } from './index';
 export const depositReceivedArgs = z
   .object({
     amount: u128,
-    asset: chainflipAssetEnum.transform(toUpperCase),
+    asset: chainflipAssetEnum,
```
zoheb391, line 19: `chainflipAssetEnum` does a transform already

packages/swap/src/event-handlers/depositIgnored.ts
```diff
@@ -0,0 +1,76 @@
...
+    });
+
+    await prisma.failedSwap.create({
+      data: {
+        type: 'IGNORED',
+        reason: 'BelowMinimumDeposit',
+        swapDepositChannelId: channel.id,
+        srcChain: chain,
+        destAddress: channel.destAddress,
+        destChain: assetToChainMap[channel.destAsset],
```
jerryafr, line 70: I think we need start storing dest chain into the database as soon as possible as this mapping will not work once we have duplicated assets on different chains

jerryafr (approved):

packages/swap/src/event-handlers/index.ts
```diff
@@ -129,6 +133,10 @@ const handlers = [
           name: events[`${chain}IngressEgress`].DepositReceived,
           handler: networkDepositReceived(chain),
         },
+        {
+          name: events[`${chain}IngressEgress`].DepositReceived,
```
GabrielBuragev:
```suggestion
          name: events[`${chain}IngressEgress`].DepositIgnored,
```

packages/swap/src/event-handlers/depositIgnored.ts
zoheb391: i think we will always a prefix for assets that determine their chain. eg: `arbEth` but yeah i see your point it would be nice to have that

packages/shared/src/map.ts
```diff
@@ -0,0 +1,9 @@
+import { Asset, Chain } from './enums';
+
+export const assetToChainMap: Record<Asset, Chain> = {
```
niklasnatter, line 3: dont we already have this in `packages/shared/src/enums.ts`?

GabrielBuragev (approved):

PR #501, opened 2024-01-25
feat(sdk, swap): handle egress amount zero

zoheb391:
closes WEB-809


jerryafr (approved):

packages/swap/src/event-handlers/egressAmountZero.ts
```diff
@@ -0,0 +1,39 @@
...
+}: EventHandlerArgs): Promise<void> {
+  const { swapId } = egressAmountZeroArgs.parse(event.args);
+
+  const swap = await prisma.swap.findUniqueOrThrow({
+    where: { nativeId: swapId },
+    include: { swapDepositChannel: true },
+  });
+  const channel = swap.swapDepositChannel;
+
+  if (!channel) {
```
niklasnatter, line 24: i think this error can happen for smart-contract swaps too 😕 so we cannot throw an error here

niklasnatter: maybe we can adapt the `FailedSwap` entity to be connected to a channel or a swap? not sure if thats the best solution but it would be nice to expose the error from the api for smart contract swaps too

zoheb391: yeah good point, i missed that. its already connected to a swap channel. i think we'd have to connect a `FailedSwap` also to a `Swap` to account for those scenarios.

PR #519, opened 2024-02-02
refactor(swap): handle updated egress events  [WEB-839][WEB-847]

acdibble:


acdibble:
> keeping all the old shapes in our zod parses will be a mess to maintain eventually 😕 maybe we can switch to a new set of handlers (that remove all the cruft) and use the old ones based on the cahin version/spec? obviously not something for this pr

Yeah I considered that and I don't disagree, but there is already a sufficiently high amount of logic in these handlers that I think it makes sense to sit down and figure out the best way of handling it

packages/swap/src/event-handlers/swapEgressScheduled.ts
```diff
@@ -100,29 +106,45 @@ export default async function swapEgressScheduled({
...
+    egress = await prisma.egress.update({
       where: { nativeId_chain: { chain, nativeId } },
       data: {
         amount: swap.swapOutputAmount?.sub(egressFee.toString()),
       },
-    }),
+    });
+  }
+
+  await Promise.all([
```
GabrielBuragev, line 143:
```suggestion
  await 
```

GabrielBuragev (approved):

PR #527, opened 2024-02-05
feat(swap): handle ignored egresses [WEB-809]

acdibble:


packages/swap/prisma/schema.prisma
```diff
@@ -266,3 +283,16 @@ model ChainTracking {
...
+  specVersion     Int
+  palletIndex     Int
+  errorIndex      Int
+  name            String
+  docs            String
+  ignoredEgresses IgnoredEgress[] @ignore
+
+  @@unique([specVersion, palletIndex, errorIndex])
+  @@schema("public")
+}
```
zoheb391, line 298: oof.. is this the pattern we're following  moving forward?

packages/swap/src/routes/swap.ts
```diff
@@ -130,9 +140,24 @@ router.get(
...
+    if (failedSwap || swap?.ignoredEgress) {
       state = State.Failed;
+      if (failedSwap?.reason) {
+        error = {
+          name: failedSwap.reason,
+          message: failedSwapMessage[failedSwap.reason],
+        };
+      } else if (swap?.ignoredEgress) {
+        const stateChainError = await prisma.stateChainError.findUniqueOrThrow({
+          where: { id: swap.ignoredEgress.stateChainErrorId },
```
niklasnatter, line 154: could we include this in the query above?

packages/swap/src/event-handlers/swapEgressIgnored.ts
```diff
@@ -0,0 +1,94 @@
...
+import { u64, chainflipAssetEnum, u128, hexString } from '@/shared/parsers';
+import { Prisma } from '../client';
+import env from '../config/env';
+import { EventHandlerArgs } from './index';
+
+const swapEgressIgnoredArgs = z.object({
+  asset: chainflipAssetEnum,
+  amount: u128,
+  swapId: u64,
+  reason: z.object({
```
niklasnatter, line 13: 😵‍💫

packages/swap/src/event-handlers/swapEgressIgnored.ts
```diff
@@ -0,0 +1,94 @@
...
+  prisma: Prisma.TransactionClient,
+  blockHash: string,
+  { value }: Reason,
+) => {
+  api ??= await ApiPromise.create({
+    provider: new WsProvider(env.RPC_NODE_WSS_URL),
+  });
+
+  const historicApi = await api.at(blockHash);
+  const error = new BN(value.error.slice(2), 'hex', 'le');
```
niklasnatter, line 40: 😵‍💫

packages/swap/src/event-handlers/swapEgressIgnored.ts
```diff
@@ -0,0 +1,94 @@
...
+import { EventHandlerArgs } from './index';
+
+const swapEgressIgnoredArgs = z.object({
+  asset: chainflipAssetEnum,
+  amount: u128,
+  swapId: u64,
+  reason: z.object({
+    __kind: z.literal('Module'),
+    value: z.object({
+      error: hexString,
```
niklasnatter, line 16: 😅

packages/swap/prisma/schema.prisma
niklasnatter: i thought its a bit convoluted at first but after thinking it through i think its quite good if we can do this for different kind of errors

niklasnatter: maybe we could connect the `StateChainError` to the swap directly (without the `IgnoredEgress`)?

acdibble: > maybe we could connect the `StateChainError` to the swap directly (without the `IgnoredEgress`)?

Unfortunately not, the egress models contain the egress amount.

> oof.. is this the pattern we're following moving forward?

Yeah sadly because the state chain errors need to be looked up via rpc

niklasnatter: > Unfortunately not, the egress models contain the egress amount.

i think you are not using the `amount` of the `IgnoredEgress`, right?

packages/swap/src/routes/swap.ts
acdibble: You would think so, but it doesn't work

packages/swap/prisma/schema.prisma
acdibble: The `EgressIgnored` event has this error:
```json
{
  "asset": {
    "__kind": "Btc"
  },
  "amount": "11",
  "reason": {
    "value": {
      "error": "0x06000000",
      "index": 34
    },
    "__kind": "Module"
  },
  "swapId": "2"
}
```

We have to convert the `reason` to a human-readable error message which involves looking it up in the runtime metadata. For historic swaps, we have to ensure that we are using the runtime metadata for the block when the `EgressIgnored` event happens. The easiest way to do this is imo is to look up the error message in the metadata and save it here.

I'm open to other implementations/approaches

acdibble: > > Unfortunately not, the egress models contain the egress amount.
> 
> i think you are not using the `amount` of the `IgnoredEgress`, right?

You're right, I should be

acdibble: No I actually am, `packages/swap/src/event-handlers/swapEgressIgnored.ts`, line 96

niklasnatter: > No I actually am, `packages/swap/src/event-handlers/swapEgressIgnored.ts`, line 96

yeah you are storing it, but we dont expose it anywhere afterwards i think 🙂 so for the information we expose right now, we would not need the `IgnoredEgress` model, connecting the swap directly to a `StateChainError` would be enough

niklasnatter: so the swap could have an `error` property that we could maybe use for different errors (not only the ones caused by `IgnoredEgress`)

acdibble: I've added the `ignoredEgress` information to the response so we can generate a block explorer link, etc

acdibble: > so the swap could have an error property that we could maybe use for different errors (not only the ones caused by IgnoredEgress)

this is how it works, it takes a generic `name` and `message`

niklasnatter (approved):
need to update the status type 🙂

acdibble:
> need to update the status type 🙂

Started working on this, will def do it in a follow up. This one's phat enough

packages/swap/src/event-handlers/swapEgressIgnored.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export type SwapDepositAddressReadyEvent = z.input<
+  typeof swapEgressIgnoredArgs
+>;
+
+let api: ApiPromise | undefined;
+
+const lookupFailure = async (
+  prisma: Prisma.TransactionClient,
+  blockHash: string,
```
zoheb391, line 33: this should prob go in `common.ts` ?

jerryafr (approved):

packages/swap/src/event-handlers/swapEgressIgnored.ts
```diff
@@ -0,0 +1,100 @@
...
+
+type Reason = z.output<typeof swapEgressIgnoredArgs>['reason'];
+
+export type SwapDepositAddressReadyEvent = z.input<
+  typeof swapEgressIgnoredArgs
+>;
+
+let api: ApiPromise | undefined;
+
+const lookupFailure = async (
```
GabrielBuragev, line 31:
```suggestion
const findOrCreateStateChainError = async (
```

packages/swap/src/event-handlers/swapEgressIgnored.ts
```diff
@@ -0,0 +1,100 @@
...
+        specVersion,
+        palletIndex,
+        errorIndex,
+      },
+    },
+  });
+
+  if (failureReason) return failureReason;
+
+  const registryError = api.registry.findMetaError({
```
GabrielBuragev, line 63: will this method simply throw if an `error` definition was not found for the specified `palletIndex, error` ?

packages/swap/src/event-handlers/swapEgressIgnored.ts
acdibble: will move it in a follow up

packages/swap/src/event-handlers/swapEgressIgnored.ts
acdibble: correct

PR #530, opened 2024-02-07
feat: update response status

acdibble:


packages/swap/prisma/migrations/20240208091109_remove_unused_fields/migration.sql
```diff
@@ -0,0 +1,21 @@
+/*
+  Warnings:
+
+  - The values [EgressAmountZero] on the enum `FailedSwapReason` will be removed. If these variants are still used in the database, this will fail.
+  - You are about to drop the column `type` on the `FailedSwap` table. All the data in the column will be lost.
```
acdibble, line 5: I don't think this was ever released so we an ignore this warning

packages/swap/prisma/migrations/20240208092708_add_failed_event_info/migration.sql
```diff
@@ -0,0 +1,11 @@
+-- AlterTable
+ALTER TABLE "public"."FailedSwap" ADD COLUMN     "failedAt" TIMESTAMP(3),
+ADD COLUMN     "failedBlockIndex" TEXT;
+
+-- backfill with default values
+UPDATE "public"."FailedSwap"
+SET "failedAt" = 'epoch', "failedBlockIndex" = '0-0';
```
acdibble, line 7: Backfill existing failed swaps with dummy data

packages/swap/src/event-handlers/depositIgnoredV120.ts
```diff
@@ -0,0 +1,90 @@
...
+        .transform((o) => {
+          try {
+            return encodeBitcoinAddress(o.value, env.CHAINFLIP_NETWORK);
+          } catch {
+            return null;
+          }
+        }),
+      hexString,
+    ]),
+    reason: reasonSchema,
```
acdibble, line 36: v120 adds `reason`. I initially added this to the original handler with a default reason, but then I realized that if this field is removed, we wouldn't catch that error and would silently use the default value which may be wrong.

packages/swap/prisma/schema.prisma
```diff
@@ -137,32 +137,25 @@ model SwapFee {
...
+  reason               FailedSwapReason
   destAddress          String
   destChain            Chain
   depositAmount        Decimal             @db.Decimal(30, 0)
   srcChain             Chain
   swapDepositChannel   SwapDepositChannel? @relation(fields: [swapDepositChannelId], references: [id])
   swapDepositChannelId BigInt?
   txHash               String?
+  failedAt             DateTime
+  failedBlockIndex     String
```
acdibble, line 159: save some extra info for the end users

packages/sdk/src/swap/types.ts
```diff
@@ -54,16 +54,20 @@ export interface SwapStatusRequest {
   id: string;
 }
 
-export interface CommonStatusFields extends ChainsAndAssets {
+export interface CommonStatusFields {
+  srcAsset: Asset;
+  srcChain: Chain;
+  destAsset: Asset | undefined;
+  destChain: Chain | undefined;
```
acdibble, line 61: failed swaps with a vault origin won't have these fields

packages/swap/prisma/schema.prisma
```diff
@@ -137,32 +137,26 @@ model SwapFee {
...
 model FailedSwap {
   id                   Int                 @id @default(autoincrement())
-  type                 FailedSwapType
-  reason               FailedSwapReason?
+  reason               FailedSwapReason
   destAddress          String
   destChain            Chain
   depositAmount        Decimal             @db.Decimal(30, 0)
   srcChain             Chain
+  srcAsset             Asset
```
acdibble, line 154: need this for the vault origin

packages/sdk/src/swap/types.ts
niklasnatter: sad

niklasnatter (approved):

jerryafr (approved):

PR #531, opened 2024-02-08
feat(swap): handle new deposit received event

zoheb391:
closes WEB-838

### DIFF
--- a/packages/swap/src/event-handlers/index.ts
+++ b/packages/swap/src/event-handlers/index.ts
@@ -19,7 +19,7 @@ import swapEgressIgnored from './swapEgressIgnored';
 import swapEgressScheduled from './swapEgressScheduled';
 import swapExecuted from './swapExecuted';
 import swapScheduled from './swapScheduled';
-import { networkDepositReceived as networkDepositReceived120 } from './v120/networkDepositReceived';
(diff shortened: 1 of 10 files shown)


packages/swap/src/event-handlers/index.ts
```diff
@@ -131,7 +131,7 @@ const handlers = [
         },
         {
           name: events[`${chain}IngressEgress`].DepositReceived,
-          handler: networkDepositReceived(chain),
+          handler: networkDepositReceived,
```
acdibble, line 134: this will add the same event handler for each chain

packages/swap/src/event-handlers/networkDepositReceived.ts
acdibble, line 1: we should keep this old event handler and move this logic to a new one. we can release the swap service backend in anticipation of the mainnet upgrade

packages/swap/src/event-handlers/index.ts
zoheb391: because of the new args, the event handler does not need to know which chain it is

GabrielBuragev: I think we don't need the chain anymore since we are getting the `swapId` in the event data.

GabrielBuragev (approved):

niklasnatter (approved):

PR #542, opened 2024-02-12
feat(sdk): specify boost fee on swap deposit channel creation [WEB-833]

GabrielBuragev:


packages/shared/src/broker.ts
```diff
@@ -77,9 +78,12 @@ const requestValidators = (network: ChainflipNetwork) => ({
           }),
         )
         .optional(),
+      z.number().optional(),
     ])
-    .transform(([a, b, c, d, e]) =>
-      [a, b, c, d, transformObjToSnakeCase(e)].filter(isNotNullish),
+    .transform(([a, b, c, d, e, f]) =>
+      [a, b, c, d, transformObjToSnakeCase(e), f].map((value) =>
+        isUndefined(value) ? null : value,
```
GabrielBuragev, line 85: Broker API accepts `null` as a parameter value in the case of optional params (both the `ccmMetadata` & `boost fee` are optional). 
It also doesn't throw when provided with more parameters then expected, so that means that this approach will work and is backwards compatible.

packages/sdk/src/swap/types.ts
```diff
@@ -31,6 +31,7 @@ interface ChainsAndAssets {
 
 export interface QuoteRequest extends ChainsAndAssets {
   amount: string;
+  boostFeeBps?: number;
```
niklasnatter, line 34: i think this should be on `DepositAddressResponse`

GabrielBuragev: You will also be able to specify it in a quote request. We will use this param to calculate the boost fee amount and calculate the quote amount.
the `getQuote` implementation will come at a later point.

niklasnatter: but adding it here will add it to `QuoteResponse` which is not right 🙂 it should be in the `includedFees` there

GabrielBuragev: Ah, i see. Good point. I guess we will have to move it here once we support it inside the `getQuote` request/response. For now, i will add it to the `DepositAddressRequest`.

jerryafr (approved):

packages/sdk/src/swap/sdk.ts
```diff
@@ -145,6 +145,8 @@ export class SwapSDK {
       depositChannelId: response.id,
       depositAddress: response.depositAddress,
       brokerCommissionBps: response.brokerCommissionBps,
+      boostFeeBps:
+        response.boostFeeBps || depositAddressRequest.boostFeeBps || 0,
```
niklasnatter, line 149: should we add this to the response of `requestSwapDepositAddress` called above as well?

packages/swap/src/handlers/openSwapDepositChannel.ts
```diff
@@ -85,6 +85,7 @@ export default async function openSwapDepositChannel(
     id: `${issuedBlock}-${srcChain}-${channelId}`,
     depositAddress: channelDepositAddress,
     brokerCommissionBps,
+    boostFeeBps: input.boostFeeBps, // TODO: fetch from db
```
niklasnatter, line 88: why not do that already? 🤔

niklasnatter (approved):

packages/swap/src/handlers/openSwapDepositChannel.ts
GabrielBuragev: there is a separate ticket for that

packages/sdk/src/swap/sdk.ts
GabrielBuragev: wdym here? the rpc response doesn't include the `boost_fee` in it.

packages/swap/src/handlers/openSwapDepositChannel.ts
GabrielBuragev: I can get that done in a follow up PR.

packages/sdk/src/swap/sdk.ts
niklasnatter: okay lol then its already consistent at least 😅

niklasnatter: okay we include the input value in the `response` object we assign in the block above, thats why i thought it is returned from the broker. might make sense to change that to be consistent

packages/shared/src/guards.ts
```diff
@@ -10,6 +10,9 @@ export const isNotNull = <T>(value: T | null): value is T => value !== null;
 export const isNotNullish = <T>(value: T | null | undefined): value is T =>
   value !== null && value !== undefined;
 
+export const isUndefined = <T>(value: T | undefined): value is T =>
```
acdibble, line 13:
```suggestion
export const isUndefined = <T>(value: T | undefined): value is undefined =>
```

packages/shared/src/broker.ts
acdibble, line 85: `undefined` in an array gets converted to `null` in JSON:

```ts
console.log(JSON.stringify([undefined]));
// => '[null]'
```

```suggestion
        value,
```

acdibble (approved):

packages/shared/src/broker.ts
GabrielBuragev: Gut point

GabrielBuragev: I just removed that logic

PR #545, opened 2024-02-13
feat(swap): index and store boost fee amount [WEB-883]

GabrielBuragev:


packages/swap/src/event-handlers/swapDepositAddressReady.ts
```diff
@@ -15,6 +15,7 @@ const swapDepositAddressReadyArgs = z.object({
   brokerCommissionRate: z.number().int(),
   sourceChainExpiryBlock: u128.optional(),
   channelMetadata: ccmMetadataSchema.optional(),
+  boostFee: z.number().int(),
```
acdibble, line 18:
```suggestion
  boostFee: z.number().int().optional(),
```

GabrielBuragev: Its actually always going to be present inside the event payload.
Screenshot 2024-02-13 at 15 01 28

PR #547, opened 2024-02-14
feat: extend quote fetching logic to include fee [WEB-832]

GabrielBuragev:


packages/shared/src/schemas.ts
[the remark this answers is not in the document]
GabrielBuragev: Maybe we can postpone its merge for a while?

packages/shared/src/functions.ts
```diff
@@ -13,3 +13,8 @@ export const bigintMin = (...args: bigint[]): bigint =>
 
 export const bigintMax = (...args: bigint[]): bigint =>
   args.reduce((max, current) => (current > max ? current : max));
+
+export const calculateBpsAmountFromTotalAmount = (
```
zoheb391, line 17:
```suggestion
export const calculateBpsAmountFromAmount = (
```

packages/sdk/src/swap/types.ts
```diff
@@ -31,16 +31,16 @@ interface ChainsAndAssets {
 
 export interface QuoteRequest extends ChainsAndAssets {
   amount: string;
+  boostFeeBps?: number;
```
niklasnatter, line 34: did we forget to add `brokerCommissionBps` here?

niklasnatter: okay the broker fee is read from the config, so i guess it makes sense 🤔

packages/shared/src/functions.ts
```diff
@@ -13,3 +13,6 @@ export const bigintMin = (...args: bigint[]): bigint =>
 
 export const bigintMax = (...args: bigint[]): bigint =>
   args.reduce((max, current) => (current > max ? current : max));
+
+export const calculateBpsAmountFromAmount = (bps: bigint, amount: bigint) =>
```
niklasnatter, line 17: we have something like this in `packages/swap/src/utils/fees.ts`, maybe we can merge them

niklasnatter (approved):

packages/shared/src/functions.ts
GabrielBuragev: Good catch. I didn't see the existence of the other function.
Moved it to shared as well as the test for it.

GabrielBuragev: just realized that they are different functions. Looking into how i can merge them

niklasnatter (approved):

PR #548, opened 2024-02-14
feat(swap): v120 adjust fee calculation

zoheb391:


packages/swap/src/event-handlers/swapEgressScheduled.ts
```diff
@@ -129,7 +135,7 @@ export default async function swapEgressScheduled({
     });
   } else {
     egressFee = bigintMin(
-      await getEgressFeeAtBlock(block.hash, swap.destAsset),
+      await getEgressFeeAtBlock(block.hash, block.specId, swap.destAsset),
```
niklasnatter, line 138: i think we dont need to handle this case because with spec 120 the fee should be on the event

niklasnatter: if its not on the event, the rpc will return the old format

packages/swap/src/routes/quote.ts
```diff
@@ -57,10 +54,8 @@ const quote = (io: Server) => {
       }
 
       const includedFees: SwapFee[] = [];
-      const ingressFee = await estimateIngressEgressFeeAssetAmount(
-        await getNativeIngressFee(query.srcAsset),
-        query.srcAsset.asset,
-      );
+      const ingressFee = await getNativeIngressFee(query.srcAsset);
```
niklasnatter, line 57: `native` is not correct anymore here 🙈 think we can rename them to `getIngressFee` and `getEgressFee`

niklasnatter (commented):
looks right to me otherwise

packages/swap/src/event-handlers/swapEgressScheduled.ts
zoheb391: ok interesting i didn't realize that was the case.. based on the ticket you wrote.. it sounded like we were to read this value from the rpc 

https://linear.app/chainflip/issue/WEB-861/adjust-quoted-ingressegress-fees-to-protocol-changes

packages/swap/src/routes/__tests__/quote.test.ts
```diff
@@ -277,7 +269,8 @@ describe('server', () => {
     });
 
     it('gets the quote from usdc when the broker is best', async () => {
-      const sendSpy = jest
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const spy = jest
```
GabrielBuragev, line 273:
```suggestion
      
      jest
```

GabrielBuragev (approved):

packages/swap/src/event-handlers/swapEgressScheduled.ts
niklasnatter: yeah i was not sure at that time how the fees will be exposed from the events 🙂 but we can leave the change, it should not hurt

zoheb391: okay i fixed it. think doesn't make sense to handle it when the event change already does it 👍

niklasnatter (approved):

PR #550, opened 2024-02-16
chore: Refactor asset identifier to match chainflip state chain

niklasnatter:


packages/sdk/src/swap/assets.ts
```diff
@@ -1,62 +1,48 @@
...
-  [Assets.BTC]: 'Bitcoin',
-};
-
 const assetFactory =
-  (asset: Asset): AssetFn =>
-  (network, env) => {
-    const assetAndChain = { asset, chain: assetChains[asset] } as AssetAndChain;
-
-    return {
-      id: asset,
```
niklasnatter, line 34: removing the id here is a small bc break, but leaving the id like it is (`ETH`, `USDC`, ...) would be confusing bc it will not identify the asset anymore (there will be two `USDC` assets with arbitrum)

packages/sdk/src/swap/chains.ts
```diff
@@ -4,21 +4,21 @@ import { ChainData } from './types';
 export const ethereum: (network: ChainflipNetwork) => ChainData = (
   network,
 ) => ({
-  id: Chains.Ethereum,
+  chain: Chains.Ethereum,
```
niklasnatter, line 7: changed this to make the `ChainData` consistent to `AssetData`

packages/sdk/src/swap/services/ApiService.ts
```diff
@@ -61,7 +61,9 @@ const getQuote: BackendQuery<
   const { brokerCommissionBps, ...returnedRequestData } = quoteRequest;
   const params: QuoteQueryParams = {
```
niklasnatter, line 62: the `quote` endpoint of the api requires a `srcChain` and a `destChain` now to prepare for arbitrum (there will be two `USDC` assets with arbitrum)

niklasnatter: these were already mandator for the `getQuote` method so its not a bc break in the sdk, but it is a change in the interface of the (internal) api

packages/shared/src/enums.ts
```diff
@@ -7,6 +7,10 @@ const arrayToMap = <const T extends readonly string[]>(
 ): ArrayToMap<T> =>
   Object.fromEntries(array.map((key) => [key, key])) as ArrayToMap<T>;
 
+export const InternalAssets = arrayToMap(['Flip', 'Usdc', 'Dot', 'Eth', 'Btc']);
```
niklasnatter, line 10: i decided to call this `InternalAssets` after all because using `ChainflipAsset` felt too generic when working with it (we would have a `Asset` and a `ChainflipAsset` in the same context). i think `InternalAsset` is a bit more descriptive and also easier to change if we need to (if you depend on something that is prefixed with `Internal` you better expect changes lol)

packages/swap/src/quoting/schemas.ts
```diff
@@ -6,21 +6,21 @@ export const marketMakerResponseSchema = z.union([
     .object({
       id: z.string(),
       intermediate_amount: numericString,
-      egress_amount: numericString,
+      output_amount: numericString,
```
niklasnatter, line 9: i kind of unintentionally changed this and then figured that the old name was misleading since we have an egress fee. this should be fine as none of the market makers is connected to this system anyway

packages/shared/src/enums.ts
```diff
@@ -26,71 +30,130 @@ export type ChainflipNetwork =
...
+export const readChainAssetMap = <T>(
+  value: ChainAssetMap<T>,
+  asset: UncheckedAssetAndChain,
+): T => {
+  assertIsValidAssetAndChain(asset);
+  const chainValues = value[asset.chain];
+  return chainValues[asset.asset as keyof typeof chainValues];
+};
+
+export function getInternalAsset(asset: UncheckedAssetAndChain) {
```
acdibble, line 143:
```suggestion
export function toInternalAsset(asset: UncheckedAssetAndChain) {
```

packages/shared/src/enums.ts
```diff
@@ -26,71 +30,130 @@ export type ChainflipNetwork =
...
 
-function assertAsset(asset: string): asserts asset is Asset {
-  if (!(asset in Assets)) throw new Error('invalid asset');
+export function isValidAssetAndChain(
+  assetAndChain: UncheckedAssetAndChain,
+): assetAndChain is AssetAndChain {
+  const { asset, chain } = assetAndChain;
+  const validAssets = chainConstants[chain].assets as string[];
+
+  return validAssets.includes(asset);
```
acdibble, line 123: This isn't a safe cast because we haven't validated that `chain` is actually a chain
```suggestion
  if (!isValidChain(chain)) return false;
  const validAssets = chainConstants[chain].assets;

  return validAssets.includes(asset);
```

packages/swap/prisma/migrations/20240216112846_change_asset_values/migration.sql
```diff
@@ -0,0 +1,8 @@
+-- AlterEnum
+BEGIN;
```
acdibble, line 2:
```suggestion
```

packages/swap/prisma/migrations/20240216112846_change_asset_values/migration.sql
```diff
@@ -0,0 +1,8 @@
...
+ALTER TYPE "public"."Asset" RENAME VALUE 'FLIP' TO 'Flip';
+ALTER TYPE "public"."Asset" RENAME VALUE 'USDC' TO 'Usdc';
+ALTER TYPE "public"."Asset" RENAME VALUE 'DOT' TO 'Dot';
+ALTER TYPE "public"."Asset" RENAME VALUE 'ETH' TO 'Eth';
+ALTER TYPE "public"."Asset" RENAME VALUE 'BTC' TO 'Btc';
+COMMIT;
```
acdibble, line 8:
```suggestion
```

packages/swap/prisma/migrations/20240216113200_rename_chain_asset_enum/migration.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
```
acdibble, line 1:
```suggestion
```

packages/swap/prisma/migrations/20240216113200_rename_chain_asset_enum/migration.sql
```diff
@@ -0,0 +1,3 @@
...
+ALTER TYPE "public"."Asset" RENAME TO "InternalAsset";
+COMMIT;
```
acdibble, line 3:
```suggestion
```

packages/sdk/src/swap/services/ApiService.ts
niklasnatter: implemented in https://github.com/chainflip-io/chainflip-sdk-monorepo/pull/550/commits/54ae86c1f33bdce4510331094b80ae7f384c4a1e

packages/swap/prisma/migrations/20240216112846_change_asset_values/migration.sql
niklasnatter: is this just for performance or is there a problem with doing it in a tx?

packages/shared/src/enums.ts
zoheb391: i think i prefer get (as it's the first keyword i think of when searching)not a big fan of using `to..` unless its a chained method

niklasnatter: no strong opinion about this, ill do whatever you can agree on

packages/shared/src/enums.ts
niklasnatter: not sure about this - the `chain` already has the type `Chain` here because 
```
UncheckedAssetAndChain = { asset: Asset; chain: Chain };
```
should i add an additional check?

packages/swap/src/routes/quote.ts
```diff
@@ -110,41 +113,43 @@ const quote = (io: Server) => {
...
-        const { id = undefined, ...response } = {
+
+        // remove the id and output amount from the final response body as it is internal information
+        const {
+          id = undefined, // eslint-disable-line @typescript-eslint/no-unused-vars
+          outputAmount,
+          ...response
+        } = {
           ...bestQuote,
           egressAmount: egressAmount.toString(),
```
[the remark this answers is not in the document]
niklasnatter: i would prefer not changing it if we dont have a important reason because its probably one of the types every consumer depends on. changing the prop name would mean all integrators need to change their integration

jerryafr (approved):

packages/shared/src/enums.ts
```diff
@@ -26,71 +30,130 @@ export type ChainflipNetwork =
...
-  assertAsset(asset);
-  if (chain !== assetChains[asset]) {
+  if (!isValidAssetAndChain(assetAndChain)) {
     throw new Error('invalid asset and chain combination');
   }
 }
+
+export const readChainAssetMap = <T>(
+  value: ChainAssetMap<T>,
+  asset: UncheckedAssetAndChain,
```
jerryafr, line 136:
```suggestion
  assetAndChain: UncheckedAssetAndChain,
```

packages/shared/src/enums.ts
```diff
@@ -26,71 +30,130 @@ export type ChainflipNetwork =
...
 ): asserts assetAndChain is AssetAndChain {
-  const { asset, chain } = assetAndChain;
-  assertAsset(asset);
-  if (chain !== assetChains[asset]) {
+  if (!isValidAssetAndChain(assetAndChain)) {
     throw new Error('invalid asset and chain combination');
   }
 }
+
+export const readChainAssetMap = <T>(
```
GabrielBuragev, line 134:
```suggestion
export const readChainAssetValue = <T>(
```

packages/shared/src/enums.ts
```diff
@@ -26,71 +30,130 @@ export type ChainflipNetwork =
...
+  value: ChainAssetMap<T>,
```
GabrielBuragev, line 135:
```suggestion
  map: ChainAssetMap<T>,
```

packages/shared/src/enums.ts
```diff
@@ -26,71 +30,130 @@ export type ChainflipNetwork =
...
+    contractId: 4,
+  },
+  [InternalAssets.Btc]: {
+    chain: Chains.Bitcoin,
+    asset: Assets.BTC,
+    name: 'Bitcoin',
+    decimals: 8,
+    contractId: 5,
+  },
+} satisfies Record<
```
GabrielBuragev, line 70:
```suggestion
} as const satisfies Record<
```

packages/shared/src/enums.ts
```diff
@@ -26,71 +30,130 @@ export type ChainflipNetwork =
...
+    assets: [Assets.DOT],
+    gasAsset: Assets.DOT,
+    contractId: 2,
+  },
+  [Chains.Bitcoin]: {
+    assets: [Assets.BTC],
+    gasAsset: Assets.BTC,
+    contractId: 3,
+  },
+} satisfies Record<
```
GabrielBuragev, line 98:
```suggestion
} as const satisfies Record<
```

packages/shared/src/enums.ts
acdibble: Yeah I guess the type is wrong then. We do use this to parse and validate user input. Maybe we should change the types there to be `string` instead, but we definitely need a check here

packages/shared/src/enums.ts
acdibble: I don't care too much. It's just weird to me to have it be called `get` when it's a constant operation and every path could be hard-coded. To me it's more of a conversion than a lookup.

packages/swap/prisma/migrations/20240216112846_change_asset_values/migration.sql
GabrielBuragev: i think its good to leave them here in case something goes wrong mid migration

packages/swap/src/routes/quote.ts
```diff
@@ -43,9 +56,11 @@ const quote = (io: Server) => {
       }
 
       const query = queryResult.data;
+      const srcAssetChain = { asset: query.srcAsset, chain: query.srcChain };
```
GabrielBuragev, line 59:
```suggestion
      const srcChainAsset = { asset: query.srcAsset, chain: query.srcChain };
```

packages/swap/src/routes/quote.ts
```diff
@@ -43,9 +56,11 @@ const quote = (io: Server) => {
...
+      const destAssetChain = { asset: query.destAsset, chain: query.destChain };
```
GabrielBuragev, line 60:
```suggestion
      const destChainAsset = { asset: query.destAsset, chain: query.destChain };
```

packages/swap/prisma/migrations/20240216112846_change_asset_values/migration.sql
acdibble: migrations are already wrapped in transactions by prisma

GabrielBuragev (approved):
Well done!

packages/sdk/src/swap/assets.ts
niklasnatter: Extracted into https://github.com/chainflip-io/chainflip-sdk-monorepo/pull/557

packages/sdk/src/swap/chains.ts
niklasnatter: Extracted into https://github.com/chainflip-io/chainflip-sdk-monorepo/pull/557

packages/swap/src/routes/quote.ts
```diff
@@ -110,41 +125,43 @@ const quote = (io: Server) => {
         const bestQuote = findBestQuote(marketMakerQuotes, brokerQuote);
 
         const quoteSwapFees = await calculateIncludedSwapFees(
-          quoteRequest.source_asset,
-          quoteRequest.destination_asset,
-          quoteRequest.deposit_amount,
+          getInternalAsset(srcAssetChain),
+          getInternalAsset(destAssetChain),
+          String(swapInputAmount),
```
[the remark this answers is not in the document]
niklasnatter: no `quoteRequest.deposit_amount` was set to `String(swapInputAmount)`. it was just a confusing property name (after we introduced the ingress fee)

packages/shared/src/enums.ts
niklasnatter: changed the type to `{ asset: string; chain: string };`

PR #570, opened 2024-02-26
feat(swap, sdk): Add liquidity warning and price impact warning 

zoheb391:
[WEB-920]


packages/shared/src/schemas.ts
```diff
@@ -62,10 +62,17 @@ export type SwapFee = {
   amount: string;
 };
 
+export type PriceWarning = {
+  threshold: number;
+  warn: boolean;
+};
+
```
zoheb391, line 69: open to renaming/changing this object

packages/shared/src/schemas.ts
```diff
@@ -62,10 +62,17 @@ export type SwapFee = {
...
 export type QuoteQueryResponse = {
   intermediateAmount?: string;
   egressAmount: string;
   includedFees: SwapFee[];
+  lowLiquidityWarning: PriceWarning;
+  priceImpactWarning: PriceWarning;
```
niklasnatter, line 75: we need to think about these names to prevent people from missunderstanding them. eg `price impact` is a term that is used by uniswap for something else 🙈  i guess we should as explicit at possible

packages/swap/src/pricing/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  },
+  BTC: {
+    chainId: 'btc',
+    address: '0x0000000000000000000000000000000000000000',
+  },
+};
+
+const priceCache = new CacheMap<string, Promise<number | undefined>>(10_000);
+
+export const GET_TOKEN_PRICE = gql(/* GraphQL */ `
```
niklasnatter, line 40: i am not sure if we should load the price from our (closed source) cache here bc we want people to be able to run the sdk without depending on our services. i think loading it from coingecko directly would be more appropriate

packages/swap/src/utils/dataStructures.ts
```diff
@@ -17,7 +17,7 @@ class Timer<T> {
   }
 
   getValue(): T {
-    return this.reset().value;
```
niklasnatter, line 20: i think the `reset` is necessary to refresh the timeout (and keep the cache value alive)

packages/shared/src/schemas.ts
niklasnatter: maybe `feeRatioWarning`/`highFeeRatioWarning` if the fees are a substantial amount of the ingress value?

niklasnatter: `lowLiquidityWarning` is a bit misleading too because this would also happen if we have enough liquidity (but at a bad price). maybe `lowSwapRateWarning`?

niklasnatter: the other possibility would be not exposing a warning but the actual value that triggers the warning - then the consumer should decide about the thresholds. in this case we could call it:
- `estimatedFeeRatio`
- `estimatedPriceDelta`

packages/swap/src/utils/dataStructures.ts
zoheb391: The issue with having a reset here.. means that the cache will continue to hit if we have a ton of traffic.  Its not right for every case. In the processor we already cache by block specific attributes with high timeouts so its not really an issue.

jerryafr (approved):

packages/swap/src/routes/quote.ts
```diff
@@ -113,6 +114,12 @@ const quote = (io: Server) => {
         );
 
         const bestQuote = findBestQuote(marketMakerQuotes, brokerQuote);
+        const lowLiquidityWarning = await checkPriceWarning({
+          srcAsset: query.srcAsset.asset, // TODO: use internal asset
```
niklasnatter, line 118: you could change this now when rebasing

packages/swap/src/utils/dataStructures.ts
niklasnatter: i see, makes sense in this case. hope we dont break any other place where we use the `CacheMap`

niklasnatter (approved):
looks good! sorry for being a pain

PR #576, opened 2024-02-29
feat(swap,sdk): [WEB-684] read witness safety margin from rpc and expose it

jerryafr:
[WEB-684]

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@chainflip/sdk",
-  "version": "1.2.1-alpha.1",
+  "version": "1.2.1-alpha.2",
   "description": "",
   "exports": {
     "./swap": {
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@chainflip/sdk",
-  "version": "1.2.1-alpha.2",
+  "version": "1.2.1-alpha.1",
   "description": "",
   "exports": {
     "./swap": {
--- a/packages/sdk/src/swap/chains.ts
+++ b/packages/sdk/src/swap/chains.ts
@@ -1,5 +1,6 @@
 import { ChainflipNetwork, Chains, isTestnet } from '@/shared/enums';
 import { ChainData } from './types';
+import { isNotNullish } from '../guards';
 import { Environment } from '../rpc';
 
 export const ethereum: (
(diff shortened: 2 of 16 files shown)


packages/sdk/src/swap/types.ts
```diff
@@ -77,6 +77,7 @@ interface DepositAddressFields extends SwapStatusResponseCommonFields {
   estimatedDepositChannelExpiryTime: number;
   isDepositChannelExpired: boolean;
   depositChannelOpenedThroughBackend: boolean;
+  witnessSafetyMargin: number | undefined;
```
niklasnatter, line 80: i am not sure if we should add this here. i think the minimum amount is kind of the same type of info so maybe we should handle them the same?

niklasnatter: minimum amount is on the asset data object if i remember correctly, so this would be on the chain data object?

jerryafr: that's right, updated it

niklasnatter (approved):
not sure if `witnessSafetyMargin` is a very intuitive name but im also not sure if there is a better alternative. so at least this is consistent to the backend 😅

packages/sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@chainflip/sdk",
-  "version": "1.2.1-alpha.1",
+  "version": "1.2.1-alpha.2",
```
niklasnatter, line 3: think you can reset this, well release it in a separate pr once its merged 🙂

packages/sdk/src/swap/sdk.ts
```diff
@@ -218,4 +223,16 @@ export class SwapSDK {
 
     return { minimumSwapAmounts: minimumDepositAmounts, maximumSwapAmounts };
   }
+
+  async getRequiredBlockConfirmations(
```
niklasnatter, line 227: i think i would return a `ChainMap` here because we return a map from `getSwapLimits` too. makes sense to keep it consistent if there is no reason against it

niklasnatter (approved):
🥳

packages/sdk/src/swap/chains.ts
```diff
@@ -1,24 +1,46 @@
...
+  env: Pick<Environment, 'ingressEgress'>,
+) => ChainData = (network, env) => ({
   chain: Chains.Bitcoin,
   name: 'Bitcoin',
   isMainnet: !isTestnet(network),
+  requiredBlockConfirmations: isNotNullish(
+    env.ingressEgress.witnessSafetyMargins.Bitcoin,
+  )
+    ? Number(env.ingressEgress.witnessSafetyMargins.Bitcoin) + 1
+    : undefined,
```
GabrielBuragev, line 45: Could use a factory method here same like in the `assets.ts` module ?

GabrielBuragev (approved):

PR #577, opened 2024-03-01
chore(swap): add logging on quote response

GabrielBuragev:


jerryafr (approved):

niklasnatter (approved):
should we also log the request?

GabrielBuragev:
> should we also log the request?

hmm, should we output the whole request object? That might be too much

PR #580, opened 2024-03-01
chore: Export getInternalAsset method from published packages

niklasnatter:


packages/sdk/src/swap/index.ts
```diff
@@ -8,3 +8,4 @@ export {
   Assets,
   ChainflipNetworks,
 } from '@/shared/enums';
+export { getInternalAsset as getChainflipId } from '@/shared/enums';
```
niklasnatter, line 11: the only place where we expose the internal asset id from the sdk is the return value of the `getAssets()` method. the property is called `chainflipId` there, so it makes sense to keep that term consistent in the `@chainflip/sdk` package

niklasnatter: exposing this method is helpful if you need a unique id for an asset-chain pair (like we do in our swapping app)

PR #595, opened 2024-03-06
chore(swap): Include ingress transaction inclusion in estimated swap duration

niklasnatter:


packages/swap/src/utils/swap.ts
```diff
@@ -6,13 +6,24 @@ export const estimateSwapDuration = async (
...
+  // user transaction must be included before witnessing starts
+  const ingressTransactionDuration = chainConstants[srcChain].blockTimeSeconds;
+
+  // once transaction is included, state chain validator witness transaction after safety margin is met
+  const ingressWitnessDuration =
     chainConstants[srcChain].blockTimeSeconds *
     Number((await getWitnessSafetyMargin(srcChain)) ?? 1n);
+
+  // once ingress is witnessed, swap and egress are executed on the state chain
   const swapDuration = CHAINFLIP_STATECHAIN_BLOCK_TIME_SECONDS * 3;
```
GabrielBuragev, line 18: Are we already accounting for the 2 block swap delay here or is there another reason why we are multiplying by 3 here?

GabrielBuragev (approved):

packages/swap/src/utils/swap.ts
niklasnatter: i am not perfectly sure, the value came from the protocol team. but i think its an estimation for how long it takes to do the swap, request the egress and sign the broadcast. so the delay would not be included

PR #596, opened 2024-03-06
feat: Add USDT asset for private test networks

niklasnatter:


packages/shared/src/enums.ts
```diff
@@ -53,6 +60,13 @@ export const assetConstants = {
     decimals: 6,
     contractId: 3,
   },
+  [InternalAssets.Usdt]: {
+    chain: Chains.Ethereum,
+    asset: Assets.USDT,
+    name: 'USDT',
+    decimals: 6,
+    contractId: 6, // TODO: update this once the final contract id is merged into the backend repo
```
niklasnatter, line 68: nee to update this once the assets can be found in the backend: https://github.com/chainflip-io/chainflip-backend/blob/4540356c923f30777aea46446b81066db1203a6d/state-chain/primitives/src/chains/assets.rs#L402

packages/shared/src/vault/schemas.ts
```diff
@@ -14,6 +14,7 @@ const bytesToHex = (arr: Uint8Array | number[]) =>
 
 const utf8ToHex = (str: string) => `0x${Buffer.from(str).toString('hex')}`;
 
+// TODO: replace zod base validation with more readable custom validation
```
niklasnatter, line 17: we generate a zod schema here to validate all possible combinations for ccm params. this is not very readable and forces the consumer to cast (https://github.com/chainflip-io/chainflip-frontend-monorepo/blob/95bcb524a9352c2e87a47f0e17084e9288808615/packages/swap/integrations/chainflip.ts#L653-L656)

niklasnatter: for now it should be okay to just add usdt to the exising schema. but the schema will grow a lot when we add the arbitrum assets, so i would suggest refactoring this before we work on arbitrum

packages/shared/src/rpc/index.ts
```diff
@@ -97,16 +97,21 @@ export const getFundingEnvironment = createRequest(
...
+  defaultValue: z.input<Z>,
+) =>
   z.object({
     Bitcoin: z.object({ BTC: parser }),
-    Ethereum: z.object({ ETH: parser, USDC: parser, FLIP: parser }),
+    Ethereum: z.object({
+      ETH: parser,
+      USDC: parser,
+      FLIP: parser,
+      USDT: parser.default(defaultValue), // remove default once usdt is available in all networks
```
jerryafr, line 110:
```suggestion
      USDT: parser.default(defaultValue), // TODO: remove default once usdt is available in all networks
```

GabrielBuragev (approved):

PR #607, opened 2024-03-11
feat(swap): add monitoring to quoting client

zoheb391:
[WEB-964]

### DIFF
--- a/packages/swap/src/__tests__/integration.test.ts
+++ b/packages/swap/src/__tests__/integration.test.ts
@@ -142,6 +142,7 @@ describe('python integration test', () => {
       id: "doesn't matter",
       intermediateAmount: '2000000000',
       outputAmount: '0', // this shouldn't be the result
+      quoteType: 'broker',
     });
 
     const response = await fetch(`${serverUrl}/quote?${params.toString()}`);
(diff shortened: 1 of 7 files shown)


packages/swap/src/quoting/QuotingClient.ts
```diff
@@ -48,13 +48,31 @@ export default class QuotingClient extends EventEmitter {
...
+      console.log('connect_error', err.message);
+      // retry with a new signature
+      const updatedTimestamp = Date.now();
+      this.socket.auth = {
+        timestamp: updatedTimestamp,
+        client_version: '1',
+        market_maker_id: this.marketMakerId,
+        signature: await this.getSignature(updatedTimestamp),
+      };
+    });
```
zoheb391, line 75: this is missing in the python client. If anyone is keen to write some python ;)

GabrielBuragev (approved):

packages/swap/src/__tests__/integration.test.ts
```diff
@@ -142,6 +142,7 @@ describe('python integration test', () => {
       id: "doesn't matter",
       intermediateAmount: '2000000000',
       outputAmount: '0', // this shouldn't be the result
+      quoteType: 'broker',
```
niklasnatter, line 145: do we need to strip this from the response of the `/quote` endpoint? not sure if we want to expose it there (yet)

niklasnatter: okay i think you already did that 🙂

niklasnatter (approved):

PR #608, opened 2024-03-12
fix(swap): handle usdc to usdc swap fees

jerryafr:


packages/swap/src/utils/fees.ts
```diff
@@ -26,7 +26,10 @@ export const calculateIncludedSwapFees = async (
   const networkFeeHundredthPips = getPoolsNetworkFeeHundredthPips(
     env.CHAINFLIP_NETWORK,
   );
-  const pools = await getPools(srcAsset, destAsset);
```
niklasnatter, line 29: could we just return an empty array if both are usdc? bc there are no fees technically

jerryafr: so changed it to return network fee only

zoheb391 (approved):

packages/swap/src/utils/fees.ts
```diff
@@ -40,10 +50,12 @@ export const calculateIncludedSwapFees = async (
         type: 'LIQUIDITY',
         chain: assetConstants[srcAsset].chain,
         asset: assetConstants[srcAsset].asset,
-        amount: getPips(
-          swapInputAmount,
-          pools[0].liquidityFeeHundredthPips,
-        ).toString(),
+        amount: pools
+          ? getPips(
```
niklasnatter, line 54: dont need to check this anymore

niklasnatter (approved):

PR #610, opened 2024-03-12
chore: Do not restrict valid asset inputs for RPC calls

niklasnatter:


packages/shared/src/broker.ts
```diff
@@ -65,8 +65,8 @@ const rpcResult = z.union([
 const requestValidators = (network: ChainflipNetwork) => ({
   requestSwapDepositAddress: z
     .tuple([
-      assetAndChain,
-      assetAndChain,
+      uncheckedAssetAndChain,
```
niklasnatter, line 68: restricting the valid values here prevents us from using this helper function for assets that are not added to the sdk yet (like the arbitrum assets right now). this forces consumers to implement the logic by themselves if they want to test new assets: https://github.com/chainflip-io/chainflip-backend/pull/4630/files#diff-01ffc1b2717a544e8ff61faa93c4d2d5bdf769fa9c1da4c8c50aeadc88863f1aR23

niklasnatter: i think its fine to allow any value here because the broker/rpc will return an error anyway if it does not know the given asset. so this just replaces the validation error with an rpc error in the failure case

zoheb391 (approved):

jerryafr (approved):

GabrielBuragev (approved):

PR #623, opened 2024-03-14
feat: Prepare swap sdk for supporting multiple evm chains [WEB-870]

niklasnatter:


packages/sdk/src/swap/sdk.ts
```diff
@@ -159,46 +157,70 @@ export class SwapSDK {
...
   }
 
   async executeSwap(
     params: ExecuteSwapParams,
-    txOpts: TransactionOptions = {},
+    txOpts: TransactionOptions & { signer?: Signer } = {},
   ): Promise<TransactionHash> {
     const { srcChain, srcAsset, amount } = params;
 
+    const { signer: optsSigner, ...remainingTxOpts } = txOpts;
```
niklasnatter, line 173: allow to pass signer when calling contract interaction methods. this allows to pass the right signer for the source chain of a swap

packages/shared/src/__tests__/contracts.test.ts
```diff
@@ -37,7 +37,7 @@ describe(checkAllowance, () => {
 
     const result = await checkAllowance(spend, spender, '0x0', signer);
 
-    expect(result.isAllowable).toBe(expected);
+    expect(result.hasSufficientAllowance).toBe(expected);
```
niklasnatter, line 40: i was always confused by `isAllowable` 😅 this is not exposed anywhere so we can change it

packages/sdk/src/swap/sdk.ts
```diff
@@ -44,44 +45,41 @@ export type SwapSDKOptions = {
...
-  private readonly rpcConfig: RpcConfig;
-
-  constructor(options: SwapSDKOptions = {}) {
-    this.network = options.network ?? ChainflipNetworks.perseverance;
-    this.baseUrl = options.backendUrl ?? BACKEND_SERVICE_URLS[this.network];
-    this.signer = options.signer;
-    this.brokerConfig = options.broker;
+  constructor(options: Partial<SwapSDKOptions> = {}) {
+    const network = options.network ?? ChainflipNetworks.perseverance;
+    this.options = {
```
niklasnatter, line 58: make options handling consistent to `FundingSDK`

packages/sdk/src/swap/sdk.ts
```diff
@@ -159,46 +157,70 @@ export class SwapSDK {
...
+    txOpts: TransactionOptions & { signer?: Signer } = {},
   ): Promise<TransactionHash> {
     const { srcChain, srcAsset, amount } = params;
 
+    const { signer: optsSigner, ...remainingTxOpts } = txOpts;
+    const signer = optsSigner ?? this.options.signer;
+    assert(signer, 'No signer provided');
+    assert(signer.provider, 'Signer has no provider');
+
+    const sourceChainId = getEvmChainId(params.srcChain, this.options.network);
```
niklasnatter, line 178: check that given signer is on the correct chain for the swap, this will be important once we support multiple evm chains

jerryafr (approved):

packages/sdk/src/swap/chains.ts
```diff
@@ -10,17 +11,17 @@ type ChainFn = (
...
-      isMainnet: !isTestnet(network),
-      requiredBlockConfirmations: isNotNullish(
-        env.ingressEgress.witnessSafetyMargins[Chains[chain]],
-      )
-        ? Number(env.ingressEgress.witnessSafetyMargins[Chains[chain]]) + 1
-        : undefined,
-    }) as ChainData;
+  (network, env) => ({
+    chain,
+    name: chain,
```
GabrielBuragev, line 16: `chain` and `name` hold the same value here. Could we merge them?

packages/sdk/src/swap/sdk.ts
```diff
@@ -33,9 +34,9 @@ import type {
 type TransactionHash = `0x${string}`;
 
 export type SwapSDKOptions = {
-  network?: ChainflipNetwork;
+  network: ChainflipNetwork;
+  backendUrl: string;
```
GabrielBuragev, line 38: i think this should be optional.

GabrielBuragev: Why are both not optional anymore?

GabrielBuragev: Ah, i see you declared the options in the constructor as `Partial`. Hmm. I guess that will work for the time being but once we add a new required property here, that will be set as Optional in the constructor type and someone might forget to change this. 
Can we somehow split the types of both `options` objects in a better way?

packages/sdk/src/swap/chains.ts
niklasnatter: not easily because removing one of the properties would be a breaking change in the interface of the sdk 🙈

niklasnatter: but i think its okay to have them, eg it would allow us to add arbitrum with `chain: 'Arbitrum'` and `name: 'Arbitrum One'` (which is the networks actual name i think)

packages/sdk/src/swap/sdk.ts
```diff
@@ -159,46 +157,70 @@ export class SwapSDK {
...
+    const signer = optsSigner ?? this.options.signer;
+    assert(signer, 'No signer provided');
+    assert(signer.provider, 'Signer has no provider');
+
+    const sourceChainId = getEvmChainId(params.srcChain, this.options.network);
+    const { chainId: signerChainId } = await signer.provider.getNetwork();
+    assert(
+      signerChainId === BigInt(sourceChainId ?? -1),
+      `Signer is connected to unexpected evm chain (expected: ${sourceChainId}, got: ${signerChainId})`,
+    );
```
GabrielBuragev, line 215: Could we dedup this block? I see its identical to the one in `executeSwap` fn.

packages/shared/src/stateChainGateway/approval.ts
```diff
@@ -36,10 +36,10 @@ export const approveStateChainGateway = async (
   networkOpts: FundingNetworkOptions,
   txOpts: TransactionOptions,
 ): Promise<ContractTransactionReceipt | null> => {
-  const { allowance, erc20, isAllowable } =
+  const { allowance, erc20, hasSufficientAllowance } =
     await checkStateChainGatewayAllowance(amount, networkOpts);
 
-  if (isAllowable) return null;
```
GabrielBuragev, line 42: uf

packages/sdk/src/swap/chains.ts
GabrielBuragev: Ah okay, i thought these values are just used internally. Makes sense then

GabrielBuragev: > but i think its okay to have them, eg it would allow us to add arbitrum with `chain: 'Arbitrum'` and `name: 'Arbitrum One'` (which is the networks actual name i think)

Indeed!

packages/sdk/src/swap/sdk.ts
niklasnatter: added a `Required` helper to make properties of the class variable non nullable

packages/sdk/src/swap/sdk.ts
niklasnatter: extracted a `assertSignerIsConnectedToChain` util and moved the check down the call stack

PR #624, opened 2024-03-15
chore: Refactor executeSwap parameter validation [WEB-870]

niklasnatter:
test swaps on sisyphos:
Screenshot 2024-03-21 at 14 54 35

### DIFF
--- a/packages/shared/src/vault/__tests__/executeSwap.test.ts
+++ b/packages/shared/src/vault/__tests__/executeSwap.test.ts
@@ -331,6 +331,28 @@ describe(executeSwap, () => {
     );
   });
 
+  it('rejects if source asset and destination asset is the same', async () => {
+    await expect(
+      executeSwap(
+        {
+          amount: '1',
+          destAsset: Assets.ETH,
+          destChain: Chains.Ethereum,
+          destAddress: ETH_ADDRESS,
+          srcAsset: Assets.ETH,
+          srcChain: Chains.Ethereum,
+        },
+        {
+          network: ChainflipNetworks.sisyphos,
+          signer: new VoidSigner('MY ADDRESS').connect({
+            getNetwork: () => Promise.resolve({ chainId: 11155111n }),
+          } as any),
+        },
+        { nonce: 1 },
+      ),
+    ).rejects.toThrow('source asset and destination asset cannot be the same');
+  });
+
   it('rejects an invalid destination address', async () => {
     await expect(
       executeSwap(
--- a/packages/shared/src/vault/executeSwap.ts
+++ b/packages/shared/src/vault/executeSwap.ts
@@ -36,6 +36,22 @@ const encodeAddress = (chain: Chain, address: string) => {
   throw new Error(`cannot encode address for chain ${chain}`);
 };
 
+export const assertIsValidSwap = (params: ExecuteSwapParams) => {
+  const internalSrcAsset = getInternalAsset({
+    chain: params.srcChain,
+    asset: params.srcAsset,
+  });
+  const internalDestAsset = getInternalAsset({
+    chain: params.destChain,
+    asset: params.destAsset,
+  });
+
+  assert(
+    internalSrcAsset !== internalDestAsset,
+    `source asset and destination asset cannot be the same`,
+  );
+};
+
 const getVaultContract = (chain: Chain, networkOpts: SwapNetworkOptions) => {
   const vaultContractAddress =
     networkOpts.network === 'localnet'
@@ -207,6 +223,7 @@ const executeSwap = async (
   networkOpts: SwapNetworkOptions,
   txOpts: TransactionOptions,
 ): Promise<ContractTransactionReceipt> => {
+  assertIsValidSwap(params);
   assertIsEvmChain(params.srcChain);
   assertValidAddress(params.destChain, params.destAddress, networkOpts.network);
   await assertSignerIsConnectedToChain(networkOpts, params.srcChain);


packages/shared/src/contracts.ts
```diff
@@ -66,9 +72,15 @@ export const approve = async (
   return transaction.wait(txOpts.wait);
 };
 
-export const getVaultManagerContractAddress = (
+export const getVaultContractAddress = (
+  chain: Chain,
```
niklasnatter, line 76: prepare for multiple chains with a vault contract

packages/shared/src/vault/executeSwap.ts
```diff
@@ -191,25 +183,26 @@ const callToken = async (
 };
 
 const executeSwap = async (
```
niklasnatter, line 185: TODO: port over useful tests from `packages/shared/src/vault/__tests__/schemas.test.ts`

packages/shared/src/vault/executeSwap.ts
```diff
@@ -1,58 +1,89 @@
...
+} from '../enums';
+import { assertIsEvmChain, assertSignerIsConnectedToChain } from '../evm';
+import { assert } from '../guards';
+import { dotAddress } from '../parsers';
+import { ccmMetadataSchema } from '../schemas';
+import { Required } from '../types';
+import { assertValidAddress } from '../validation/addressValidation';
+import { ExecuteSwapParams, SwapNetworkOptions } from './index';
+
+const encodeAddress = (chain: Chain, address: string) => {
```
niklasnatter, line 28: TODO: ensure this is called in a test

packages/shared/src/parsers.ts
```diff
@@ -44,14 +44,24 @@ export const hexStringFromNumber = numericString.transform(
...
 );
 
-export const btcAddress = (network: ChainflipNetwork) => {
+export const btcAddress = (network: ChainflipNetwork | 'localnet') => {
   if (network === 'mainnet') {
-    return string.regex(/^(1|3|bc1)/).refine(validateBitcoinMainnetAddress);
+    return string
+      .regex(/^(1|3|bc1)/)
+      .refine(validateBitcoinMainnetAddress, (address) => ({
+        message: `${address} is not a valid bitcoin address`,
```
niklasnatter, line 52: make address validation error messages a bit nicer than 
```
      {
        "code": "custom",
        "message": "Invalid input",
        "path": [
          "address"
        ]
      }
```

packages/shared/src/vault/__tests__/__snapshots__/executeSwap.test.ts.snap
```diff
@@ -19,7 +19,7 @@ exports[`executeSwap can be invoked with localnet options 1`] = `
 [
   [
     3,
-    "0x74623171676539767664326d6d6a7866687578757132303468346678787068723076666e736e78323035",
+    "0x626372743170376d6e6c6c36717570346c336c6367677668377432366d3467727961776a793064306367616e7a683665347a6a6d326433727471743975737178",
```
niklasnatter, line 22: snapshot changed bc i needed to pass a different btc address (we previously passed a testnet address, but it needs to be a regtest on backspin)

packages/shared/src/parsers.ts
```diff
@@ -44,14 +44,24 @@ export const hexStringFromNumber = numericString.transform(
...
   return z.union([
-    string.regex(/^(m|n|2|tb1)/).refine(validateBitcoinTestnetAddress),
-    string.regex(/^bcrt1/).refine(validateBitcoinRegtestAddress),
+    string
+      .regex(/^(m|n|2|tb1)/)
+      .refine(validateBitcoinTestnetAddress, (address) => ({
+        message: `${address} is not a valid bitcoin address`,
+      })),
+    string.regex(/^bcrt1/).refine(validateBitcoinRegtestAddress, (address) => ({
+      message: `${address} is not a valid bitcoin address`,
```
acdibble, line 63:
```suggestion
      message: `"${address}" is not a valid Bitcoin regtest address`,
```

packages/shared/src/validation/addressValidation.ts
```diff
@@ -131,10 +134,25 @@ const validators: Record<ChainflipNetwork, Record<Chain, AddressValidator>> = {
...
 ): boolean => validators[network][chain](address);
+
+export const assertValidAddress = (
+  chain: Chain,
+  address: string,
+  network: ChainflipNetwork | 'localnet',
+) =>
+  assert(
+    validators[network][chain](address),
+    `Address "${address}" is not a valid ${chain}" address`,
```
acdibble, line 157:
```suggestion
    `Address "${address}" is not a valid "${chain}" address`,
```

packages/shared/src/vault/__tests__/executeSwap.test.ts
```diff
@@ -274,13 +276,154 @@ describe(executeSwap, () => {
...
+          } as any),
+        },
+        { nonce: 1 },
+      ),
+    ).rejects.toThrow(
+      'Signer is connected to unexpected evm chain (expected: 11155111, got: 404)',
+    );
+  });
+
+  it('rejects if the destination chain is not an evm chain for a call', async () => {
```
GabrielBuragev, line 404: we can only do ccm between evm chains?

packages/shared/src/vault/executeSwap.ts
```diff
@@ -149,35 +173,23 @@ const callToken = async (
...
-      ? networkOpts.vaultContractAddress
-      : getVaultManagerContractAddress(networkOpts.network);
-
-  const erc20Address =
-    networkOpts.network === 'localnet'
-      ? networkOpts.srcTokenContractAddress
-      : getTokenContractAddress(internalSrcAsset, networkOpts.network);
-
-  assert(erc20Address !== undefined, 'Missing ERC20 contract address');
+  const { vaultContract: vault, vaultAddress } = getVaultContract(
```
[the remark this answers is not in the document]
niklasnatter: the `getInternalAsset` call will error if the pair is not a valid pair 🙂 but i will add an explicit check to make it more obvious

packages/shared/src/vault/__tests__/executeSwap.test.ts
niklasnatter: at the moment, yes. i think that we will support it for solana in the future, but i dont think that was implemented yet. i tried to verify and this is the event you see when you try to do a ccm swap to bitcoin:
Screenshot 2024-03-21 at 15 31 54

packages/shared/src/vault/executeSwap.ts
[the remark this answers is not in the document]
acdibble: `getInternalAsset` will error if the chain and asset don't match. But we don't ever check that the two assets are compatible

packages/shared/src/vault/__tests__/executeSwap.test.ts
niklasnatter: this should be the place where the backend defines if a chain supports a ccm swap: https://github.com/chainflip-io/chainflip-backend/blob/b1f4da7641b0d66fb98c708df88fbbf5072c33d1/state-chain/primitives/src/chains.rs#L122

packages/shared/src/vault/executeSwap.ts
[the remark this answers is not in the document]
niklasnatter: oh i see, i misunderstood the word pair 🙂 how do we define a valid pair though?

GabrielBuragev (approved):

packages/shared/src/vault/__tests__/executeSwap.test.ts
GabrielBuragev: Good to know! Thanks!

packages/shared/src/vault/executeSwap.ts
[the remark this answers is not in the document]
niklasnatter: i guess its invalid if it is from/to the same asset? i just did a swap from and to `Eth` and it seems like the protocol accepted it, so im not sure if we should allow it or not 🙈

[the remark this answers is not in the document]
niklasnatter: https://scan.sisyphos.staging/swaps/19 😵‍💫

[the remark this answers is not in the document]
niklasnatter: added an `assertIsValidSwap` to assert that the source and destination assets are not the same

packages/shared/src/parsers.ts
```diff
@@ -72,10 +82,13 @@ export const dotAddress = z
...
     }
   })
-  .refine(isString);
+  .refine(isString, {
+    message: `address is not a valid polkadot address`,
+  });
 
-export const ethereumAddress = hexString.refine((address) =>
-  ethers.isAddress(address),
+export const ethereumAddress = hexString.refine(
```
zoheb391, line 89: 🙌

zoheb391 (approved):

packages/cli/src/commands/cliExecuteSwap.ts
```diff
@@ -82,7 +82,9 @@ export default async function cliExecuteSwap(
           args.chainflipNetwork,
         );
 
-  const wallet = new Wallet(privateKey).connect(getDefaultProvider(ethNetwork));
+  const wallet = new Wallet(privateKey).connect(
+    getDefaultProvider(ethNetwork, { etherscan: '-' }), // disable etherscan provider because it is unreliable
```
GabrielBuragev, line 86: What was unreliable about it?

packages/shared/src/vault/executeSwap.ts
```diff
@@ -1,58 +1,116 @@
...
+  );
+};
+
+const getVaultContract = (chain: Chain, networkOpts: SwapNetworkOptions) => {
   const vaultContractAddress =
     networkOpts.network === 'localnet'
       ? networkOpts.vaultContractAddress
-      : getVaultManagerContractAddress(networkOpts.network);
+      : getVaultContractAddress(chain, networkOpts.network);
+
```
GabrielBuragev, line 60: Should we also assert here that the address is defined?

packages/shared/src/evm.ts
```diff
@@ -1,8 +1,13 @@
 import { Signer } from 'ethers';
 import { getEvmChainId } from './consts';
-import { Chain, ChainflipNetwork } from './enums';
+import { Chain, ChainflipNetwork, ChainflipNetworks } from './enums';
 import { assert } from './guards';
 
+export const assertIsEvmChain = (chain: Chain) => {
+  const evmChainId = getEvmChainId(chain, ChainflipNetworks.backspin);
```
GabrielBuragev, line 7: Are we always supposed to check against `backspin` network?

GabrielBuragev (approved):

packages/shared/src/vault/executeSwap.ts
niklasnatter: the `getVaultContractAddress` will throw if its not defined

packages/shared/src/evm.ts
niklasnatter: this kinda a shortcut to prevent explicitly adding the information if a chain is an evm chain. if the chain (eg `Arbitrum`) has an evm chain id on backspin, we can assume that it is an evm chain (on all networks). we dont care about what the id is, only if it is set

PR #630, opened 2024-03-18
chore(swap): Handle rpc returning ingress/egress fee in fee asset instead of gas asset [WEB-924]

niklasnatter:


packages/swap/src/routes/quote.ts
```diff
@@ -63,6 +63,13 @@ const quote = (io: Server) => {
 
       logger.info('received a quote request', { query: req.query });
 
+      // detect if ingress and egress fees are exposed as gas asset amount or fee asset amount
+      // https://github.com/chainflip-io/chainflip-backend/pull/4497
+      // TODO: remove this once all networks are upraded to 1.3
+      const ingressEgressFeeIsGasAssetAmount =
+        (await getIngressFee({ chain: 'Ethereum', asset: 'FLIP' })) ===
```
niklasnatter, line 70: the `cf_environment` call inside of `getIngressFee` i cached, so this should not cause any additional requests

packages/swap/src/routes/quote.ts
```diff
@@ -63,6 +63,13 @@ const quote = (io: Server) => {
 
       logger.info('received a quote request', { query: req.query });
 
+      // detect if ingress and egress fees are exposed as gas asset amount or fee asset amount
+      // https://github.com/chainflip-io/chainflip-backend/pull/4497
+      // TODO: remove this once all networks are upraded to 1.3
+      const ingressEgressFeeIsGasAssetAmount =
```
niklasnatter, line 69: this works because `FLIP` and `USDC` ave a different number of decimals. so if the ingress fee is the same for both of them, we can be sure that it is the old version of the rpc (the value is a gas asset amount, not a FLIP/USDC amount)

jerryafr: Nice trick!

jerryafr (approved):

PR #633, opened 2024-03-19
fix: Make values in ingress_fees and egress_fees map nullable

niklasnatter:
The values are null if the fee cannot be calculated by the backend since https://github.com/chainflip-io/chainflip-backend/pull/4497/files#diff-da69d849254aec2e3c9755396682d6302755145e827801c8d0bbccad22c25c85R269

### DIFF
--- a/packages/shared/src/rpc/index.ts
+++ b/packages/shared/src/rpc/index.ts
@@ -146,8 +146,8 @@ const rename =
 const ingressEgressEnvironment = z
   .object({
     minimum_deposit_amounts: chainAssetMapFactory(numberOrHex, 0),
-    ingress_fees: chainAssetMapFactory(numberOrHex, 0),
-    egress_fees: chainAssetMapFactory(numberOrHex, 0),
+    ingress_fees: chainAssetMapFactory(numberOrHex.nullable(), null),
+    egress_fees: chainAssetMapFactory(numberOrHex.nullable(), null),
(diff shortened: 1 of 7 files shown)


packages/shared/src/rpc/index.ts
```diff
@@ -146,8 +146,8 @@ const rename =
 const ingressEgressEnvironment = z
   .object({
     minimum_deposit_amounts: chainAssetMapFactory(numberOrHex, 0),
-    ingress_fees: chainAssetMapFactory(numberOrHex, 0),
-    egress_fees: chainAssetMapFactory(numberOrHex, 0),
+    ingress_fees: chainAssetMapFactory(numberOrHex.nullable(), null),
```
acdibble, line 149: why not default it to `0`?

niklasnatter: i think we want to throw an error in the `quote` endpoint if we dont know the fee (this should only happen if the pool has no liquidity anyway). when using null i can check against that in the quote endpoint to detect it 🙂

niklasnatter: it will be null if the backend cannot estimate it, i havent looked into when exactly that happens tbh. but i think its better to be on the safe side and not return a quote instead of returning a wrong one (without the fee) in this case

acdibble (approved):

jerryafr (approved):

PR #636, opened 2024-03-20
feat: extend quote fetching logic to include fee

GabrielBuragev:


acdibble (approved):

packages/shared/src/functions.ts
```diff
@@ -13,3 +13,17 @@ export const bigintMin = (...args: bigint[]): bigint =>
 
 export const bigintMax = (...args: bigint[]): bigint =>
   args.reduce((max, current) => (current > max ? current : max));
+
+export const ONE_IN_PIP = 10000;
+export const ONE_IN_HUNDREDTH_PIPS = ONE_IN_PIP * 100;
+
+export const calculatePipAmountFromAmount = (
```
jerryafr, line 20:
```suggestion
export const getPipAmountFromAmount = (
```
just a suggestion, you can ignore if you don't like it

jerryafr (approved):

PR #637, opened 2024-03-20
fix(sdk): remove btc hash prefix, update field name

acdibble:


scripts/reportDeprecations.mts
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env node --import=tsx --trace-uncaught --no-warnings
```
niklasnatter, line 1: nice idea!

niklasnatter (approved):

jerryafr (approved):

PR #638, opened 2024-03-21
refactor(sdk): change undefined to null

acdibble:


niklasnatter (approved):
how should we decide what to use when in the future? 🤔

acdibble:
> how should we decide what to use when in the future? 🤔

image

Here at least there is precedent. I think it's important to use `null` here because looking up a key in an object that doesn't exist will return `undefined`. `null` indicates the lack of a meaningful value, whereas `undefined` would represent that this key doesn't exist.

PR #641, opened 2024-03-22
chore: Update USDT address on sepolia

niklasnatter:


packages/shared/src/consts.ts
```diff
@@ -27,7 +27,7 @@ export const getEvmChainId = (chain: Chain, network: ChainflipNetwork) => {
 const SEPOLIA_USDC_CONTRACT_ADDRESS =
   '0x1c7D4B196Cb0C7B01d743Fbc6116a902379C7238';
 const SEPOLIA_USDT_CONTRACT_ADDRESS =
```
niklasnatter, line 29: we decided to deploy our own USDT contract to sepolia to make sure tht the contract matches the mainnet one

PR #648, opened 2024-03-25
feat(swap): add tx ref to broadcast

zoheb391:
closes WEB-987

### DIFF
--- a/packages/swap/src/event-handlers/networkBroadcastSuccess.ts
+++ b/packages/swap/src/event-handlers/networkBroadcastSuccess.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import { strip0x } from '@/shared/bitcoin';
 import { Chain } from '@/shared/enums';
 import { hexString, unsignedInteger } from '@/shared/parsers';
 import type { EventHandlerArgs } from './index';
@@ -27,7 +28,10 @@ export default function networkBroadcastSuccess(
       data: {
         succeededAt: new Date(block.timestamp),
         succeededBlockIndex: `${block.height}-${event.indexInBlock}`,
-        transactionRef: args.transactionRef,
+        transactionRef:
+          chain === 'Bitcoin'
+            ? strip0x(args.transactionRef)
+            : args.transactionRef,
       },
     });
   };
--- a/packages/shared/src/bitcoin.ts
+++ b/packages/shared/src/bitcoin.ts
@@ -24,3 +24,6 @@ export const encodeAddress = (
     1,
     Buffer.from(address.slice(2), 'hex'),
   ).encode();
+
+export const strip0x = (string: string | undefined) =>
+  string?.startsWith('0x') ? string.slice(2) : string;


packages/swap/src/event-handlers/networkBroadcastSuccess.ts
```diff
@@ -1,24 +1,40 @@
...
-import { unsignedInteger } from '@/shared/parsers';
+import { hexString, unsignedInteger } from '@/shared/parsers';
 import type { EventHandlerArgs } from './index';
 
-const eventArgs = z.object({
+export const networkBroadcastSuccessArgs = z.object({
   broadcastId: unsignedInteger,
+  transactionRef: z // v130+
+    .union([
+      z.object({ blockNumber: z.number(), extrinsicIndex: z.number() }),
```
acdibble, line 10:
```suggestion
      z.object({ blockNumber: z.number(), extrinsicIndex: z.number() })
        .transform(({ blockNumber, extrinsicIndex }) => `${blockNumber}-${extrinsicIndex}`),
```

packages/swap/src/event-handlers/networkBroadcastSuccess.ts
```diff
@@ -1,24 +1,40 @@
...
+    const args = networkBroadcastSuccessArgs.parse(event.args);
+
+    let txRef;
+    if (args.transactionRef) {
+      if (typeof args.transactionRef === 'string') {
+        txRef = args.transactionRef;
+      } else {
+        txRef = `${args.transactionRef.blockNumber}-${args.transactionRef.extrinsicIndex}`;
+      }
+    }
```
acdibble, line 29:
```suggestion
```

packages/swap/src/event-handlers/networkBroadcastSuccess.ts
```diff
@@ -1,24 +1,40 @@
...
 
     // use updateMany to skip update if broadcast does not include any swap
     await prisma.broadcast.updateMany({
-      where: { chain, nativeId: broadcastId },
+      where: { chain, nativeId: args.broadcastId },
       data: {
         succeededAt: new Date(block.timestamp),
         succeededBlockIndex: `${block.height}-${event.indexInBlock}`,
+        transactionRef: txRef,
```
acdibble, line 37:
```suggestion
        transactionRef: args.transactionRef,
```

zoheb391:
We also need to expose this on the swapping backend/sdk. Do we have any preferences for what we want to call this field?

I'm thinking `outboundTxRef`
or 
`outboundTxLink` with the full block explorer link

acdibble (approved):

packages/swap/src/event-handlers/networkBroadcastSuccess.ts
zoheb391: duh ofcourse

niklasnatter (approved):
i would call the property on the status `broadcastTransactionRef` because we have most of the properties prefixed with the state chain term. and i think providing the ref (not an explorer link) is preferable bc its easier to create an explorer link from a ref than the other way

PR #649, opened 2024-03-25
feat(cli,shared,sdk): [WEB-881] add Arbitrum support

jerryafr:
[WEB-881]


packages/shared/src/node-apis/redis.ts
```diff
@@ -34,6 +34,7 @@ const broadcastParsers = {
   }),
   Polkadot: z.object({ tx_out_id: z.object({ signature: string }) }),
   Bitcoin: z.object({ tx_out_id: z.object({ hash: string }) }),
+  Arbitrum: z.object({ tx_out_id: z.object({ hash: string }) }).nullish(), // TODO: remove once arbitrum is available in all networks
```
zoheb391, line 37:
```suggestion
  Arbitrum: z.object({ tx_out_id: z.object({ hash: string }) }).nullish(), // TODO: remove once arbitrum is available on all networks
```

packages/shared/src/rpc/index.ts
```diff
@@ -90,15 +90,19 @@ const chainAssetMapFactory = <Z extends z.ZodTypeAny>(parser: Z, defaultValue: z
       USDT: parser.default(defaultValue), // TODO: remove default once usdt is available in all networks
     }),
     Polkadot: z.object({ DOT: parser }),
+    Arbitrum: z
+      .object({ ETH: parser, USDC: parser })
+      .default({ ETH: defaultValue, USDC: defaultValue }), // TODO: remove default once arbitrum is available in all networks
```
zoheb391:
```suggestion
      .default({ ETH: defaultValue, USDC: defaultValue }), // TODO: remove default once arbitrum is available on all networks
```

zoheb391 (approved):

packages/sdk/src/swap/services/ApiService.ts
```diff
@@ -52,6 +52,9 @@ const getAssets = async (
   if (chain === Chains.Bitcoin) {
     return [btc$(network, env)];
   }
+  if (chain === Chains.Arbitrum) {
```
niklasnatter, line 55: i think we want to return this for `backspin` only for now

packages/sdk/src/swap/services/__tests__/ApiService.test.ts
```diff
@@ -19,31 +19,37 @@ const env = {
       },
       Polkadot: { DOT: 0x77359400n },
       Bitcoin: { BTC: 0x5f370n },
+      Arbitrum: { ETH: 0n, USDC: 0n },
     },
     ingressFees: {
       Ethereum: { ETH: 0n, FLIP: 0n, USDC: 0n, USDT: 0n },
       Polkadot: { DOT: 0n },
       Bitcoin: { BTC: 0n },
+      Arbitrum: { ETH: 0n, USDC: 0n },
```
niklasnatter, line 28: might make sense to have one of these not set to test the mainnet case 🙂

packages/sdk/src/swap/services/ApiService.ts
niklasnatter: also we want to add arbitrum the the return value of `getChains`/`getPossibleDestinationChains`

packages/shared/src/rpc/index.ts
```diff
@@ -150,6 +154,9 @@ const rpcAsset = z.union([
   z.object({ chain: z.literal('Ethereum'), asset: z.literal('ETH') }),
   z.literal('USDC'),
   z.object({ chain: z.literal('Ethereum'), asset: z.literal('USDC') }),
+  z.object({ chain: z.literal('Ethereum'), asset: z.literal('USDT') }),
```
niklasnatter, line 157: oof another one 😕  i think its hard to enforce all assets via type here, can we generate the `{chain,asset}` variants based on our constants?

packages/swap/src/utils/rpc.ts
```diff
@@ -25,7 +25,8 @@ export const getMinimumEgressAmount = async (asset: UncheckedAssetAndChain): Pro
   return readChainAssetValue(environment.ingressEgress.minimumEgressAmounts, asset);
 };
 
-export const getWitnessSafetyMargin = async (chain: Chain): Promise<bigint | null> => {
+export const getWitnessSafetyMargin = async (chain: Chain): Promise<bigint | null | undefined> => {
+  // TODO: remove undefined once arbitrum is available on all networks
```
niklasnatter, line 29: we could just set the default value to `null`, right?

packages/shared/src/rpc/index.ts
jerryafr: tried it but couldn't make it work right now to generate them :(
problem seems to be with how zod treats arrays

packages/sdk/src/swap/services/ApiService.ts
```diff
@@ -52,6 +52,11 @@ const getAssets = async (
...
+    return network === ChainflipNetworks.backspin
+      ? [arbeth$(network, env), arbusdc$(network, env)]
```
niklasnatter, line 57: we need to return `Arbitrum` from `getChains`/`getPossibleDestinationChains`

packages/shared/src/rpc/index.ts
```diff
@@ -87,18 +87,22 @@ const chainAssetMapFactory = <Z extends z.ZodTypeAny>(parser: Z, defaultValue: z
...
+      USDT: parser.default(defaultValue), // TODO: remove default once usdt is available on all networks
     }),
     Polkadot: z.object({ DOT: parser }),
+    Arbitrum: z
+      .object({ ETH: parser, USDC: parser })
+      .default({ ETH: defaultValue, USDC: defaultValue }), // TODO: remove default once arbitrum is available on all networks
   });
 
-const chainMap = <Z extends z.ZodTypeAny>(parser: Z) =>
+const chainMap = <Z extends z.ZodTypeAny>(parser: Z, defaultValue: z.input<Z>) =>
```
niklasnatter, line 98:
```suggestion
const chainMapFactory = <Z extends z.ZodTypeAny>(parser: Z, defaultValue: z.input<Z>) =>
```

packages/shared/src/rpc/index.ts
niklasnatter: okay too bad 😕

packages/sdk/src/swap/services/__tests__/ApiService.test.ts
```diff
@@ -73,20 +80,33 @@ describe('ApiService', () => {
...
+        ).toMatchSnapshot();
       },
     );
 
     it('gets mainnet chains', async () => {
-      expect(await ApiService.getChains(ChainflipNetworks.mainnet, env)).toMatchSnapshot();
+      expect(
+        await ApiService.getChains(
+          ChainflipNetworks.mainnet,
+          env as unknown as Pick<Environment, 'ingressEgress'>,
```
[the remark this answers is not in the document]
jerryafr: in the zod object, there is a default value for arbitrum in case it's not provided, but the inferred typing still considers it as required

packages/shared/src/rpc/index.ts
jerryafr: all done, please check again 🙏

packages/sdk/src/swap/services/__tests__/ApiService.test.ts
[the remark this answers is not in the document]
niklasnatter: hm okay so the default value would have already been applied before we pass it to the `ApiService.getChains()`, right?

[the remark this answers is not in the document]
niklasnatter: i think we should add the value in the param here too then. im sorry, i think you changed this because of me 🙈

packages/sdk/src/swap/services/ApiService.ts
```diff
@@ -16,21 +16,31 @@ import {
...
   }
   if (sourceChain === Chains.Polkadot) {
-    return [ethereum(network, env), bitcoin(network, env)];
+    chains = [ethereum(network, env), bitcoin(network, env)];
   }
   if (sourceChain === Chains.Bitcoin) {
-    return [ethereum(network, env), polkadot(network, env)];
+    chains = [ethereum(network, env), polkadot(network, env)];
+  }
+  if (sourceChain === Chains.Arbitrum) {
```
niklasnatter, line 39: could check `ChainflipNetworks.backspin` here too, but no blocker for me

niklasnatter (approved):

PR #653, opened 2024-03-26
refactor(swap): fetch all asset prices at once

acdibble:


packages/swap/src/pricing/index.ts
```diff
@@ -1,58 +1,97 @@
...
+
+  shouldRefetch() {
+    return Date.now() - this.lastCacheSet < PriceCache.TTL;
   }
 
-  const { promise, resolve } = deferredPromise<number | undefined>();
+  async getAssetPrice(asset: InternalAsset): Promise<number | undefined> {
+    logger.debug(`getting asset price for "${asset}"`);
+
+    if (this.shouldRefetch()) {
```
jerryafr, line 45: I think it should not be called shouldRefetch, but isUpToDate or something like that?

acdibble: Yep, renamed

jerryafr (approved):

packages/swap/src/pricing/index.ts
```diff
@@ -1,58 +1,97 @@
...
+
+  freshEnough() {
+    return Date.now() - this.lastCacheSet < PriceCache.TTL;
   }
 
-  const { promise, resolve } = deferredPromise<number | undefined>();
+  async getAssetPrice(asset: InternalAsset): Promise<number | undefined> {
+    logger.debug(`getting asset price for "${asset}"`);
+
+    if (this.freshEnough()) {
```
zoheb391, line 45: image

packages/swap/src/pricing/index.ts
```diff
@@ -1,58 +1,97 @@
 import axios from 'axios';
 import { InternalAsset } from '@/shared/enums';
 import env from '../config/env';
-import { CacheMap } from '../utils/dataStructures';
```
niklasnatter, line 4: we removed something from the `CacheMap` to make it work for this usecase, so maybe we should readd that?
https://github.com/chainflip-io/chainflip-sdk-monorepo/pull/570/files#diff-b0ae927d6867edde94cb13c8bf6099e9f97f6e27905f178c4e481a70ec814e5dL12

acdibble: done

packages/swap/src/pricing/index.ts
```diff
@@ -1,58 +1,97 @@
...
 
 const coingeckoAxios = axios.create({
   baseURL: 'https://pro-api.coingecko.com/api/v3',
   timeout: 5000,
   headers: { 'x-cg-pro-api-key': env.COINGECKO_API_KEY },
 });
 
-export const getAssetPrice = async (asset: InternalAsset): Promise<number | undefined> => {
-  logger.debug(`getting asset price for "${asset}"`);
+export class PriceCache {
```
acdibble, line 29: using a class for encapsulation is easier than trying to reset global variables by resetting mocks in jest

PR #662, opened 2024-03-29
chore(swap): extend address screening process

GabrielBuragev:


packages/swap/src/utils/screenAddress.ts
```diff
@@ -28,3 +28,16 @@ export default async function screenAddress(address: string): Promise<boolean> {
 
   return result.data.identifications.length > 0;
 }
+
+export const addressBlacklist: string[] = [];
```
GabrielBuragev, line 32: Still waiting for the addresses.

acdibble: Make it an env var

GabrielBuragev:
I guess we can close this.

PR #676, opened 2024-04-05
refactor(swap): do not log liquidity issues

acdibble:


zoheb391 (approved):

packages/swap/src/routes/quote.ts
```diff
@@ -226,7 +226,9 @@ const quote = (io: Server) => {
         const message =
           err instanceof Error ? err.message : 'unknown error (possibly no liquidity)';
 
-        logger.error('error while collecting quotes:', err);
+        if (!message.includes('InsufficientLiquidity')) {
+          logger.error('error while collecting quotes:', err);
+        }
```
zoheb391, line 231: maybe as a warning and a have a seperate alert to warn JIT in case it goes over our threshold?

jerryafr (approved):
maybe logging InsufficientLiquidity as warning just for the sake of having more info?

packages/swap/src/routes/quote.ts
```diff
@@ -226,7 +226,9 @@ const quote = (io: Server) => {
...
 
         // DEPRECATED(1.3): remove `error`
         res.status(500).json({ message, error: message });
```
jerryafr, line 234: Also I think returning 500 would cause the result to be logged as error, but not sure

PR #677, opened 2024-04-05
chore(sdk): Allow to not wait for transaction inclusion before hash is returned

niklasnatter:


packages/shared/src/stateChainGateway/index.ts
```diff
@@ -51,23 +51,25 @@ export const fundStateChainAccount = async (
     amount,
     extractOverrides(txOpts),
   );
+  await transaction.wait(txOpts.wait);
 
-  return transaction.wait(txOpts.wait) as Promise<ContractTransactionReceipt>;
```
niklasnatter, line 55: `transaction.wait()` will return null if `txOpts.wait` is 0, because transaction receipts are only available after the tx was included

niklasnatter: instead of returning the receipt, we can return the transaction response which will be available even before the tx is included in a block

niklasnatter: the tx response will already include the tx hash (which is the only thing we use in the sdk). we can change these methods because they are not exposed from the sdk directly

CHANGELOG.md
```diff
@@ -15,6 +15,14 @@ which it is deprecated.
   string `error` field. This field has been replaced by `message` to be inline
   with how errors are returned in other parts of the API.
 
+## 1.3.1
```
acdibble, line 18: unless we're releasing it immediately
```suggestion
## Unreleased
```

acdibble (approved):

PR #692, opened 2024-04-08
chore(swap): handle deposit received event name change [WEB-1010]

GabrielBuragev:


packages/swap/src/event-handlers/index.ts
```diff
@@ -55,14 +57,16 @@ export const events = {
     EgressScheduled: 'EthereumIngressEgress.EgressScheduled',
     BatchBroadcastRequested: 'EthereumIngressEgress.BatchBroadcastRequested',
     CcmBroadcastRequested: 'EthereumIngressEgress.CcmBroadcastRequested',
-    DepositReceived: 'EthereumIngressEgress.DepositReceived',
+    DepositReceived: 'BitcoinIngressEgress.DepositReceived', // Renamed to DepositFinalised since spec 140
+    DepositFinalised: 'BitcoinIngressEgress.DepositFinalised',
```
niklasnatter, line 61: arent you adding the `DepositFinalised` in the `spec: 140` block below too?

niklasnatter: oh im stupid, can forget about this 🙈

packages/swap/src/event-handlers/index.ts
```diff
@@ -193,6 +198,13 @@ const handlers = [
       ]),
     ],
   },
+  {
+    spec: 140,
+    handlers: Object.values(Chains).map((chain) => ({
+      name: events[`${chain}IngressEgress`].DepositFinalised,
+      handler: depositFinalised,
```
zoheb391, line 205: can we add tests for this new event handler?

GabrielBuragev: Yeah, i wanted to do that in a follow up PR so we unbrick the deposits on bspin.

zoheb391: isn't it just a copy paste of the other file?

GabrielBuragev: Oh, i thought there is no test for `depositReceived` as i couldn't find it in the `v120` folder. In that case i'll copy and adjust that test here.

GabrielBuragev: Ready sir

zoheb391 (approved):

PR #700, opened 2024-04-17
fix(swap): fix build

acdibble:


jerryafr (approved):
I'm wondering if the following commands are still correct or not 🤔 
```
COPY --from=builder /app/packages/swap /app/packages/swap
COPY --from=builder /app/packages/amm-addon /app/packages/amm-addon
```

acdibble:
> I'm wondering if the following commands are still correct or not 🤔 
> 
> ```
> 
> COPY --from=builder /app/packages/swap /app/packages/swap
> 
> COPY --from=builder /app/packages/amm-addon /app/packages/amm-addon
> 
> ```

Yeah that should be the dist folder

PR #706, opened 2024-04-21
chore(swap): Support DISABLED_INTERNAL_ASSETS environment variable [WEB-1065]

niklasnatter:


packages/swap/src/config/env.ts
```diff
@@ -45,6 +45,7 @@ export default z
     USE_JIT_QUOTING: optionalBoolean,
     QUOTE_APPROXIMATION_THRESHOLD: optionalNumber(0.1),
     STEALTH_MODE: optionalBoolean.default('true'),
+    DISABLED_INTERNAL_ASSETS: z.string().transform((s) => s.split(',')), // required to make sure that it is easily editable in argocd
```
niklasnatter, line 48: TODO: add this to all deployments

packages/swap/src/handlers/openSwapDepositChannel.ts
```diff
@@ -24,6 +24,12 @@ export default async function openSwapDepositChannel(
   }
 
   const { srcAsset, destAsset } = getInternalAssets(input);
+  if (env.DISABLED_INTERNAL_ASSETS.includes(srcAsset)) {
```
niklasnatter, line 27: TODO: add test cases

packages/swap/src/routes/quote.ts
```diff
@@ -117,15 +117,18 @@ const quoteRouter = (io: Server) => {
...
-        throw ServiceError.unavailable('Polkadot is disabled');
-      }
-
       const start = performance.now();
 
       logger.info('received a quote request', { query: req.query });
       const query = queryResult.data;
+
       const { srcAsset, destAsset } = queryResult.data;
+      if (env.DISABLED_INTERNAL_ASSETS.includes(srcAsset)) {
```
niklasnatter, line 126: TODO: add test cases

packages/swap/src/config/env.ts
```diff
@@ -45,6 +47,7 @@ export default z
     USE_JIT_QUOTING: optionalBoolean,
     QUOTE_APPROXIMATION_THRESHOLD: optionalNumber(0.1),
     STEALTH_MODE: optionalBoolean.default('true'),
+    DISABLED_INTERNAL_ASSETS: optionalString('').transform((s) => s.split(',')),
```
acdibble, line 50: Should filter non-internal assets and log a warning if we find any

packages/swap/src/config/env.ts
```diff
@@ -45,6 +48,14 @@ export default z
     USE_JIT_QUOTING: optionalBoolean,
     QUOTE_APPROXIMATION_THRESHOLD: optionalNumber(0.1),
     STEALTH_MODE: optionalBoolean.default('true'),
+    DISABLED_INTERNAL_ASSETS: optionalString('').transform((string) =>
+      string.split(',').map((asset) => {
+        if (asset && !(asset in InternalAssets)) {
+          console.warn(`unexpected value in DISABLED_INTERNAL_ASSETS variable: "${asset}"`);
```
niklasnatter, line 54: cant use logger here bc it depends on `env` being parsed

packages/swap/src/config/env.ts
```diff
@@ -45,6 +48,15 @@ export default z
     USE_JIT_QUOTING: optionalBoolean,
     QUOTE_APPROXIMATION_THRESHOLD: optionalNumber(0.1),
     STEALTH_MODE: optionalBoolean.default('true'),
+    DISABLED_INTERNAL_ASSETS: optionalString('').transform((string) =>
+      string.split(',').map((asset) => {
```
jerryafr, line 52: we can use `forEach` here I think

niklasnatter: think we cannot use foreach because we need to return the array (this is inside the transform)

packages/swap/src/config/env.ts
```diff
@@ -45,6 +48,15 @@ export default z
...
+        if (asset && !(asset in InternalAssets)) {
+          // eslint-disable-next-line no-console
+          console.warn(`unexpected value in DISABLED_INTERNAL_ASSETS variable: "${asset}"`);
```
acdibble, line 55:
```suggestion
          console.warn({ message: `unexpected value in DISABLED_INTERNAL_ASSETS variable: "${asset}"` });
```

acdibble (approved):

PR #713, opened 2024-04-23
feat(swap): add new pool table and handle event [WEB-1008]

GabrielBuragev:
~~These pools will be created with a migration. That means that they will be available on all networks without requiring an extrinsic call.~~

~~For localnets, they will be enabled by default (safe mode flags will be turned on: deposit_enabled, withdraw_enabled, boost_enabled). For public networks, they will be disabled by default and enabled via an extrinsic call (event will be emitted). We will capture this event to update the safe mode flags.~~

~~The approach i am taking is the following:~~
- ~~One off job to initialize the pools based on the environment. If localnet, set flags to true. If pubnet set flags to false.~~
- ~~On testnets, we will rely on an event to update these pools. A follow up PR will introduce the event handler for this~~

Based on the latest discussion, we can expect an event when a boost pool was created.

### DIFF
--- a/packages/swap/prisma/schema.prisma
+++ b/packages/swap/prisma/schema.prisma
@@ -276,6 +276,18 @@ model Pool {
   @@schema("public")
 }
 
+model BoostPool {
+  id              Int           @id @default(autoincrement())
+  depositEnabled  Boolean
+  withdrawEnabled Boolean
+  boostEnabled    Boolean
+  asset           InternalAsset
+  feeTierPips     Int
+
+  @@unique([asset, feeTierPips])
+  @@schema("public")
+}
+
 model ChainTracking {
   id             Int      @id @default(autoincrement())
   chain          Chain    @unique
--- a/packages/swap/src/event-handlers/v140/boostPoolCreated.ts
+++ b/packages/swap/src/event-handlers/v140/boostPoolCreated.ts
@@ -1,0 +1,39 @@
+import { z } from 'zod';
+import { assetAndChain, number } from '@/shared/parsers';
+import { screamingSnakeToPascalCase } from '@/shared/strings';
+import { EventHandlerArgs } from '..';
+import { InternalAsset } from '../../client';
+
+const boostPoolCreatedEventSchema = z.object({
+  boostPool: z.object({
+    asset: assetAndChain.transform((chainAsset) => ({
+      ...chainAsset,
+      asset: screamingSnakeToPascalCase(chainAsset.asset),
+    })),
+    tier: number,
+  }),
+});
+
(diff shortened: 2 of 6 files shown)


packages/shared/src/enums.ts
```diff
@@ -30,6 +30,9 @@ export type ChainflipNetwork = (typeof ChainflipNetworks)[keyof typeof Chainflip
 export const isTestnet = (network: ChainflipNetwork): boolean =>
   network !== ChainflipNetworks.mainnet;
 
+export const isLocalnet = (network: ChainflipNetwork): boolean =>
+  network === ChainflipNetworks.backspin || network === ChainflipNetworks.sisyphos;
```
acdibble, line 34: sisyphos should undergo the runtime upgrade
```suggestion
  network === ChainflipNetworks.backspin;
```

packages/swap/prisma/schema.prisma
```diff
@@ -276,6 +276,18 @@ model Pool {
   @@schema("public")
 }
 
+model BoostPool {
+  id              Int           @id @default(autoincrement())
+  depositEnabled  Boolean
+  withdrawEnabled Boolean
+  boostEnabled    Boolean
```
acdibble, line 283: what will we use these for? usually we just read directly from RPC when required

GabrielBuragev: Hmm, well the RPCs will not return this info. Also, I think storing these details in the db is much more efficient as we will be able to change them based on events.
Additionally, we need them broken down in separate flags so we can determine exactly which part is disabled. This will give us more power when exposing boost details via the SDK. 

Hope this makes sense to you.

niklasnatter (approved):

packages/swap/src/event-handlers/v140/boostPoolCreated.ts
```diff
@@ -0,0 +1,39 @@
...
+    tier: number,
+  }),
+});
+
+export const boostPoolCreated = async ({ prisma, event }: EventHandlerArgs) => {
+  const {
+    boostPool: { asset: chainAsset, tier },
+  } = boostPoolCreatedEventSchema.parse(event.args);
+
+  await prisma.boostPool.upsert({
```
jerryafr, line 22: should we expect to receive this event multiple times for the same asset? maybe it's better to throw error in that case 🤔

jerryafr (approved):

packages/swap/src/event-handlers/v140/boostPoolCreated.ts
GabrielBuragev: Who knows 😅 
I think its better to ignore it rather than the server crash looping

PR #721, opened 2024-05-02
feat(swap): load existing orders for quoter

acdibble:


packages/swap/src/quoting/Leg.ts
```diff
@@ -32,6 +32,14 @@ export default class Leg {
     return this.destAsset !== 'Usdc' ? 'BUY' : 'SELL';
   }
 
+  getBaseAsset(): Exclude<InternalAsset, 'Usdc'> {
+    if (this.destAsset !== 'Usdc') return this.destAsset;
+
+    if (this.srcAsset !== 'Usdc') return this.srcAsset;
+
+    throw new Error('one of the assets must be Usdc');
```
niklasnatter, line 40:
```suggestion
    throw new Error('one of the assets must not be Usdc');
```

packages/swap/src/utils/rpc.ts
```diff
@@ -39,3 +40,20 @@ export const getEgressFee = async (asset: InternalAsset): Promise<bigint | null>
 
   return readChainAssetValue(environment.ingressEgress.egressFees, asset);
 };
+
+const fetchOrdersAndPrice = async (baseAsset: Exclude<InternalAsset, 'Usdc'>) => {
+  const [orders, price] = await Promise.all([
+    getPoolOrders(rpcConfig, getAssetAndChain(baseAsset), getAssetAndChain('Usdc')),
```
niklasnatter, line 46: maybe we should give them a suffix to indicate that the result is not parsed

niklasnatter: but no strong opinion as theay are typed anyway i think?

packages/swap/src/utils/rpc.ts
```diff
@@ -39,3 +40,20 @@ export const getEgressFee = async (asset: InternalAsset): Promise<bigint | null>
...
+  return { poolState: orders, rangeOrderPrice: price.rangeOrder };
+};
+
+const ordersCacheMap = new AsyncCacheMap({
+  refresh: false,
+  refreshInterval: 6_000,
+  fetch: fetchOrdersAndPrice,
+});
+
+export const getCachedPoolOrdersAndPrice = ordersCacheMap.get.bind(ordersCacheMap);
```
niklasnatter, line 59: what does this do? 🤔

niklasnatter (approved):

packages/swap/src/quoting/Leg.ts
acdibble: One of them must be `Usdc`. We return the one that isn't

PR #722, opened 2024-05-02
refactor(swap): fetch state in loop

acdibble:


packages/swap/src/quoting/PoolStateCache.ts
```diff
@@ -0,0 +1,101 @@
...
+  private async startPolling() {
+    while (this.running) {
+      const hash = await getBlockHash(rpcConfig);
+
+      if (hash !== this.latestHash) {
+        this.latestHash = hash;
+
+        const success = await this.cacheMap.load(hash);
+
+        assert(success, 'cache should be loaded');
```
niklasnatter, line 79: i guess this will break out of the `while`, right? so we wont recover if we fail one time?

niklasnatter (approved):

packages/swap/src/quoting/PoolStateCache.ts
acdibble: Yeah wasn't sure what to do, I think I'll always catch here and then let the caller handle the error instead of a unhandled promise rejection

zoheb391 (commented):
should we also factor in block timestamp checks before returning the poolstate? i.e if difference is greater than some pre determined value - don't return pool state. Trying to think what happens when the rpc is stale

PR #730, opened 2024-05-07
feat: include boosted quote in quote response [WEB-1017]

GabrielBuragev:
adds a boosted quote inside `boostedQuote` property within the quote response.

The quote includes the normal swap estimated time which should be updated to use the boosted swap time.


packages/swap/src/utils/fees.ts
```diff
@@ -121,3 +127,38 @@ export const calculateIncludedSwapFees = async (
     },
   ];
 };
+
+export const tryExtractFeesFromIngressAmount = async ({
```
GabrielBuragev, line 131: extracted the logic for calculating the "ingress" side fees (broker + ingress) here

packages/swap/src/routes/quote.ts
```diff
@@ -21,14 +22,68 @@ import { resultify } from '../utils/promise';
...
   validateSwapAmount,
+  getBoostPoolsDepth,
 } from '../utils/rpc';
 import ServiceError from '../utils/ServiceError';
 import { estimateSwapDuration } from '../utils/swap';
 
 const getPoolQuoteResult = resultify(getPoolQuote);
 
+export const getBoostedPoolQuoteResult = async (query: ParsedQuoteParams) => {
+  const { srcAsset, destAsset, brokerCommissionBps, amount } = query;
```
GabrielBuragev, line 34: Need to add a feature flag here

packages/shared/src/rpc/index.ts
```diff
@@ -180,22 +181,27 @@ export const getIngressEgressEnvironment = createRequest(
   ingressEgressEnvironment,
 );
 
-const rpcAsset = z.union([
+const deprecatedRpcAssetSchema = z.union([
```
acdibble, line 184: not sure we even need these anymore

packages/shared/src/rpc/index.ts
```diff
@@ -268,3 +274,24 @@ export const getPoolPriceV2 = createRequest(
...
+  z.intersection(
+    rpcAssetSchema,
+    z
+      .object({
+        tier: number,
+        available_amount: u128,
+      })
+      .transform((obj) => ({
+        tier: obj.tier,
+        availableAmount: obj.available_amount,
```
acdibble, line 288: this is automatically taken care of
```suggestion
```

packages/shared/src/schemas.ts
```diff
@@ -80,11 +78,13 @@ export type SwapFee = {
...
-export type QuoteQueryResponse = {
+export type QuoteDetails = {
   intermediateAmount?: string;
   egressAmount: string;
   includedFees: SwapFee[];
   lowLiquidityWarning: boolean | undefined;
   estimatedDurationSeconds: number;
 };
+export type QuoteQueryResponse = QuoteDetails & {
+  boostInformation?: QuoteDetails & { boostFeeBps: number };
```
acdibble, line 89:
```suggestion
  boostedQuote?: QuoteDetails & { boostFeeBps: number };
```

packages/swap/src/utils/boost.ts
```diff
@@ -0,0 +1,28 @@
...
+export const getBoostFeeBpsForAmount = async ({
+  amount,
+  assetBoostPoolsDepth,
+}: {
+  amount: bigint;
+  assetBoostPoolsDepth: BoostPoolsDepth;
+}): Promise<number | undefined> => {
+  assetBoostPoolsDepth.sort((a, b) => (a.tier < b.tier ? -1 : 1));
+
+  let _amount = amount;
```
acdibble, line 13:
```suggestion
  let remainingAmount = amount;
```

packages/swap/src/routes/quote.ts
```diff
@@ -21,14 +22,68 @@ import { resultify } from '../utils/promise';
...
+    amount: BigInt(amount),
+    assetBoostPoolsDepth,
+  });
+
+  if (effectiveBoostFeeBps === undefined) return undefined;
+
+  const boostFee = getPipAmountFromAmount(amount, effectiveBoostFeeBps);
+  fees.push(buildFee(srcAsset, 'BOOST', boostFee));
+
+  if (amount <= 0n) {
```
acdibble, line 49: amount wasn't modified, is it supposed to be?

acdibble (commented):
looking good! small comments

packages/shared/src/rpc/index.ts
GabrielBuragev: We are also not using the `cf_pool_info` anywhere it seems like. Thats the only place where we were using the "deprecated" schemas. I guess thats also safe to remove.

packages/shared/src/rpc/index.ts
GabrielBuragev: Not sure why but the response properties are not being camel cased for some reason.

packages/swap/src/routes/quote.ts
GabrielBuragev: I missed that, there should be a check for this. Thanks for pointing out.

packages/shared/src/rpc/index.ts
GabrielBuragev: Ah, i see the issue. The response is an array of objects and we don't handle that properly in the `camelCaseKeys` method.

GabrielBuragev: Fixed it

packages/sdk/src/swap/types.ts
```diff
@@ -34,19 +34,19 @@ interface ChainsAndAssets {
...
 export interface QuoteResponse
-  extends Omit<QuoteRequest, 'boostFeeBps' | 'brokerCommissionBps' | 'affiliateBrokers'> {
+  extends Omit<QuoteRequest, 'brokerCommissionBps' | 'affiliateBrokers'> {
   quote: QuoteQueryResponse;
 }
 
 export interface DepositAddressRequest extends QuoteRequest {
   destAddress: string;
   ccmMetadata?: CcmMetadata;
+  boostFeeBps?: number;
```
jerryafr, line 49: shouldn't `boostFeeBps` be part of the DepositAddressResponse?

GabrielBuragev: It is under `DepositAddressRequest` because it is also used when opening a deposit channel. 
The `DepositAddressRequest` extends from `QuoteRequest` and the difference here (compared to the broker comission) is that we decided to not support passing a `boostFeeBps` as an argument when fetching a quote but instead calculate the effective boost fee automatically.

packages/swap/src/utils/boost.ts
```diff
@@ -0,0 +1,28 @@
...
+  let remainingAmount = amount;
+  let feeAmount = 0n;
+
+  for (const poolDepth of assetBoostPoolsDepth) {
+    const poolAvailableAmount = poolDepth.availableAmount;
+
+    const amountToBeUsedFromPool = bigintMin(remainingAmount, poolAvailableAmount);
+    feeAmount += getPipAmountFromAmount(amountToBeUsedFromPool, poolDepth.tier);
+    remainingAmount -= amountToBeUsedFromPool;
+  }
```
jerryafr, line 22:
```suggestion
    remainingAmount -= amountToBeUsedFromPool;
    if (remainingAmount === 0) break;
  }
```

jerryafr (approved):

packages/swap/src/routes/quote.ts
```diff
@@ -203,7 +233,11 @@ const quoteRouter = (io: Server) => {
         });
 
         if (result.success) {
-          res.json({ ...result.data.response, quoteType: undefined });
+          res.json({
+            ...result.data.response,
+            quoteType: undefined,
+            boostedQuote: isLocalnet() ? await getBoostedPoolQuoteResult(query) : undefined,
```
niklasnatter, line 239: think this will add the `boostedQuote` if we dont use the quoter?

niklasnatter: might make sense to move the `isLocalnet` check into `getBoostedPoolQuoteResult`, then we can call it multiple times

packages/swap/src/routes/quote.ts
niklasnatter: why would we need the feature flag?

packages/swap/src/routes/__tests__/__snapshots__/quote.test.ts.snap
```diff
@@ -40,6 +40,90 @@ exports[`server GET /quote gets the quote for deprecated params without the chai
 }
 `;
 
+exports[`server GET /quote gets the quote from btc with boost information 1`] = `
+{
+  "boostedQuote": {
+    "boostFeeBps": 5,
```
niklasnatter, line 46: should we call this `effectiveBoostFeeBps`?

niklasnatter (commented):
good work! two ways of quoting is making stuff a bit complicated atm, but it would be fine for me to support only the pool quotes for now (?)

packages/swap/src/routes/__tests__/__snapshots__/quote.test.ts.snap
```diff
@@ -40,6 +40,90 @@ exports[`server GET /quote gets the quote for deprecated params without the chai
 }
 `;
 
+exports[`server GET /quote gets the quote from btc with boost information 1`] = `
+{
+  "boostedQuote": {
```
niklasnatter, line 45: not sure if we use `boosted` somewhere, could also call it `boostQuote` to keep the feature name?

packages/swap/src/routes/quote.ts
```diff
@@ -21,14 +23,70 @@ import { resultify } from '../utils/promise';
...
+  try {
+    const { fees: includedFees, amountAfterFees } = await tryExtractFeesFromIngressAmount({
+      ingressAmount: amountAfterBoostFees,
+      srcAsset,
+      brokerCommissionBps,
+    });
+    const swapInputAmount = amountAfterFees;
+    fees.push(...includedFees);
+
+    const boostedPoolQuote = await getPoolQuoteResult(
```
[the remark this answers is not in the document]
GabrielBuragev: we decided to not use the quoter for the boosted quotes as we are planning to rework that part and use an RPC request as a replacement for the current quoter implementation. cc @acdibble

packages/shared/src/rpc/index.ts
```diff
@@ -268,3 +250,19 @@ export const getPoolPriceV2 = createRequest(
 );
 
 export const getBlockHash = createRequest('chain_getBlockHash', hexString);
+
+const boostPoolsDepthResponseSchema = z.array(
+  z.intersection(
```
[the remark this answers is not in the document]
GabrielBuragev: Not sure why but merging with a union is not possible 🤔

packages/swap/src/routes/quote.ts
```diff
@@ -21,14 +23,70 @@ import { resultify } from '../utils/promise';
...
+    assetBoostPoolsDepth,
+  });
+
+  if (effectiveBoostFeeBps === undefined) return undefined;
+
+  const boostFee = getPipAmountFromAmount(amount, effectiveBoostFeeBps);
+  fees.push(buildFee(srcAsset, 'BOOST', boostFee));
+  const amountAfterBoostFees = amount - boostFee;
+
+  if (amountAfterBoostFees <= 0n) {
```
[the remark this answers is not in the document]
GabrielBuragev: You are actually right, the boost fee is always relative to the deposit amount :D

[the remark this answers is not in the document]
GabrielBuragev: I will remove it so to simplify the method.

packages/swap/src/routes/quote.ts
```diff
@@ -21,14 +23,70 @@ import { resultify } from '../utils/promise';
...
+    );
+
+    if (!boostedPoolQuote.success) {
+      return undefined;
+    }
+
+    return {
+      ...boostedPoolQuote.data.response,
+      boostFeeBps: effectiveBoostFeeBps,
+      quoteType: undefined,
```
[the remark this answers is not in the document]
GabrielBuragev: I am not sure how and if we are using that anywhere but i just tried to be consistent with how we construct the normal swap quote where we are overwriting this value and setting it to `undefined` in the case of a pool swap quote. Maybe @acdibble has more context on this one?

packages/swap/src/routes/quote.ts
GabrielBuragev: Yeah, as i explained above, we will replace the current quoter with a simple RPC call so we decided to not spend too much time trying to make this work with the current quoter too.

I also kinda feel like the `isLocalnet` check here shouldn't be a responsibility of the `getBoostedPoolQuoteResult` method but instead the quote endpoint handler.

GabrielBuragev: Does that maybe provide more clearance/context?

packages/swap/src/routes/__tests__/__snapshots__/quote.test.ts.snap
GabrielBuragev: hmm, i do agree that `boostFeeBps` might not be very clear and it helps being more explicit here but calling it `effective` might also be wrong as it is only an estimate at this point. What about `boostFeeBpsEstimate` or `estimatedBoostFeeBps` ?

packages/swap/src/routes/__tests__/__snapshots__/quote.test.ts.snap
GabrielBuragev: Will do

packages/swap/src/routes/__tests__/__snapshots__/quote.test.ts.snap
GabrielBuragev: Used `estimatedBoostFeeBps`

packages/swap/src/routes/quote.ts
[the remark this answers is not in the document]
acdibble: The new quoting stuff will be in 1.4 so we only have to handle this path rn

packages/swap/src/routes/quote.ts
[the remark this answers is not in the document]
niklasnatter: okay im not sure if it should be on the response sent to the user, but its not related to this pr then

niklasnatter (approved):
🚀

packages/swap/src/routes/quote.ts
niklasnatter: okay not a strong opinion, its fine for me both ways 🙂

PR #731, opened 2024-05-07
feat(sdk): Add affiliateBrokers to DepositAddressRequest [WEB-1081]

niklasnatter:


packages/shared/src/schemas.ts
```diff
@@ -58,6 +60,13 @@ export const ccmMetadataSchema = z.object({
 
 export type CcmMetadata = z.infer<typeof ccmMetadataSchema>;
 
+export const affiliateBroker = z.object({
+  account: chainflipAddress,
+  commissionBps: number,
+});
+
+export type AffiliateBroker = z.infer<typeof affiliateBroker>;
```
acdibble, line 68: slightly better imo
```suggestion
export const affiliateBroker = z.object({
  account: chainflipAddress,
  commissionBps: number,
}).transform(({ account, commissionBps: bps }) => ({ account, bps }));

export type AffiliateBroker = z.input<typeof affiliateBroker>;
```

packages/shared/src/broker.ts
```diff
@@ -152,5 +170,8 @@ export async function requestSwapDepositAddress(
       cfParameters: undefined,
     },
     boostFeeBps,
+    opts.affiliateBrokers?.length
+      ? opts.affiliateBrokers.map(({ account, commissionBps }) => ({ account, bps: commissionBps }))
+      : undefined,
```
acdibble, line 175: if you don't accept the other suggestion
```suggestion
    opts.affiliateBrokers?.map(({ account, commissionBps }) => ({ account, bps: commissionBps })),
```

packages/shared/src/parsers.ts
```diff
@@ -84,6 +84,11 @@ export const ethereumAddress = hexString.refine(
   (address) => ({ message: `${address} is not a valid ethereum address` }),
 );
 
+export const chainflipAddress = string.refine(
+  (address) => address.startsWith('cF') && decodeAddress(address),
```
acdibble, line 88: I imagine the RPC will accept any substrate address but it doesn't hurt

acdibble:
And the changelog plz

packages/shared/src/broker.ts
```diff
@@ -134,7 +148,11 @@ const makeRpcRequest = async <T extends keyof RequestValidator & keyof ResponseV
 
 export async function requestSwapDepositAddress(
   swapRequest: NewSwapRequest,
-  opts: { url: string; commissionBps: number },
+  opts: {
+    url: string;
+    commissionBps: number;
+    affiliateBrokers?: AffiliateBroker[];
```
niklasnatter, line 154: we dont have a `broker` prefix for `commissionBps`, so we can rename this to `affiliates`

packages/shared/src/parsers.ts
niklasnatter: yeah i think allowing only formatted addresses is a bit safer

packages/sdk/src/swap/sdk.ts
```diff
@@ -161,6 +174,7 @@ export class SwapSDK {
       depositChannelId: response.id,
       depositAddress: response.depositAddress,
       brokerCommissionBps: response.brokerCommissionBps,
+      affiliateBrokers: depositAddressRequest.affiliateBrokers ?? [],
```
acdibble, line 177: I'm handling this in the swap status response
```suggestion
```

packages/sdk/src/swap/sdk.ts
```diff
@@ -153,6 +158,14 @@ export class SwapSDK {
         channelOpeningFee: result.channelOpeningFee,
       };
     } else {
+      assert(
+        !depositAddressRequest.brokerCommissionBps,
+        'Broker commission is only supported only when initializing the SDK with a brokerUrl',
+      );
```
acdibble, line 164: this is a breaking change for people who might instantiate the sdk with a commission but use our sdk. i would add an extra change to changelog saying that this will now throw an error instead of silently ignoring

acdibble (approved):
small changes

packages/sdk/src/swap/sdk.ts
niklasnatter: but this is the channel creation response?

acdibble: true i was thinking it's the swap status response 👍

packages/sdk/src/swap/sdk.ts
niklasnatter: wanted to add this but then i noticed that it wasnt possible to set a commission without providing a broker url before (you could only instantiate the sdk with a url + commission)

acdibble: cool, nvm then

jerryafr (approved):

PR #732, opened 2024-05-08
feat(swap): store affiliates [WEB-1131]

acdibble:


packages/sdk/src/swap/types.ts
```diff
@@ -80,6 +80,7 @@ interface DepositAddressFields extends SwapStatusResponseCommonFields {
   estimatedDepositChannelExpiryTime: number;
   isDepositChannelExpired: boolean;
   depositChannelOpenedThroughBackend: boolean;
+  affiliateBrokers?: { account: string; commissionBps: number }[];
```
niklasnatter, line 83: not sure if we should call this `beneficiaries` because it will also include the submitter? or maybe it wont?

niklasnatter: can talk about this in person

packages/swap/src/event-handlers/swapScheduled.ts
```diff
@@ -26,6 +26,8 @@ const swapScheduledArgs = z.object({
   origin: z.union([depositChannelSwapOrigin, vaultSwapOrigin]),
   swapType: swapTypeSchema,
   brokerCommission: u128.optional(),
+  // >= v1.4.0
```
niklasnatter, line 29: should we add a similar comment for `brokerCommission`?

packages/sdk/src/swap/types.ts
acdibble: We don't have access to the main broker, so I've renamed it to affiliates

packages/swap/src/event-handlers/swapScheduled.ts
acdibble: done

packages/swap/src/event-handlers/swapScheduled.ts
```diff
@@ -25,7 +25,10 @@ const swapScheduledArgs = z.object({
   destinationAddress: encodedAddress,
   origin: z.union([depositChannelSwapOrigin, vaultSwapOrigin]),
   swapType: swapTypeSchema,
-  brokerCommission: u128.optional(),
+  // < v1.4.0
+  brokerCommission: u128.nullish(),
+  // >= v1.4.0
+  brokerFee: u128.nullish(),
 });
```
jerryafr, line 32:
```suggestion
  brokerFee: u128.nullish(),
}).transform(({ brokerCommission, brokerFee, ...rest }) => ({
  brokerFee: brokerFee ?? brokerCommission,
  ...rest,
}));
```
just a suggestion

jerryafr (approved):

packages/swap/src/event-handlers/swapScheduled.ts
acdibble: done

packages/sdk/src/swap/types.ts
niklasnatter: hm would it be better if we include all beneficiaries (like in https://github.com/chainflip-io/chainflip-web-services/pull/1571)? i think we could get the submitter from the event, right?

packages/swap/src/routes/swap.ts
```diff
@@ -110,6 +124,9 @@ router.get(
 
       swap = swapDepositChannel.swaps.at(0);
       failedSwap = swapDepositChannel.failedSwaps.at(0);
+      if (swapDepositChannel.affiliates.length > 0) {
+        affiliateBrokers = swapDepositChannel.affiliates;
```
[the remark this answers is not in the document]
niklasnatter: if we had `beneficiaries`, the `brokerCommissionBps` would be the sum of all commissions of al beneficiaries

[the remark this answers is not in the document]
acdibble: I don't think we should include it in that field. To me, it makes more sense to have the response match the request format. It might make sense to add a new field `totalBrokerCommissionBps`

packages/sdk/src/swap/types.ts
acdibble: I think it makes more sense for the response format to match the request format. It's possible to get the submitter, but I don't know if there is a use case for including it. Odds are the person requesting a status with the SDK knows the broker used to create the deposit channel. If we get feedback that it would be useful, we can definitely include it. But at this stage I think it makes more sense to keep it simple and uniform with the request format

niklasnatter (approved):
not sure about affiliates vs beneficiaries, but lets merge it for now 🙂

acdibble:
> not sure about affiliates vs beneficiaries, but lets merge it for now 🙂

beneficiaries is just an internal concept to simplify logic on our side that I don't think will be documented anywhere. people interacting with the protocol will be expecting to see affiliates. can chat about it in stand up

PR #734, opened 2024-05-09
feat(swap): add blocklist

acdibble:


jerryafr (approved):
maybe later adding createdAt, updatedAt and isActive (or enabled) fields to the table would become useful

PR #742, opened 2024-05-17
fix(swap): use supported assets

acdibble:


jerryafr (approved):

packages/swap/src/quoting/PoolStateCache.ts
```diff
@@ -11,28 +17,34 @@ const rpcConfig = { rpcUrl: env.RPC_NODE_HTTP_URL };
...
 type PoolState = {
   poolState: string;
   rangeOrderPrice: bigint;
 };
 
-const fetchPoolState = async (hash: string) =>
-  Object.fromEntries(
+const fetchPoolState = async (hash: string) => {
+  baseAssets ??= (await getSupportedAssets(rpcConfig))
+    .filter((asset) => asset.chain !== 'Ethereum' || asset.asset !== 'USDC')
```
jerryafr, line 29:
```suggestion
    .filter((asset) => !(asset.chain === 'Ethereum' && asset.asset === 'USDC'))
```
just a preference, not sure if it's easier to read for others as well or not

acdibble: definitely easier to read

PR #746, opened 2024-05-17
chore(swap): rename exposed boost fee properties [WEB-852]

GabrielBuragev:
The boost fees are already showing up on multiple places within a different context (boost fee tolerance, estimated boost fee, effective boost fee). In order to reduce confusion and provide an easier to understand API, i renamed the channel opening method properties to use `maxBoostFeeBps` instead of `boostFeeBps` which should speak more clearly about the purpose.


packages/swap/prisma/migrations/20240517144322_rename_boost_fee_column/migration.sql
```diff
@@ -0,0 +1,9 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `boostFeeBps` on the `SwapDepositChannel` table. All the data in the column will be lost.
+
+*/
+-- AlterTable
+ALTER TABLE "public"."SwapDepositChannel" DROP COLUMN "boostFeeBps",
+ADD COLUMN     "maxBoostFeeBps" INTEGER NOT NULL DEFAULT 0;
```
GabrielBuragev, line 9: We never had created channels with `boostFeeBps` > 0

acdibble (approved):

niklasnatter (approved):

PR #752, opened 2024-05-21
fix(swap): handle boosters credited action

GabrielBuragev:


packages/swap/src/event-handlers/v120/networkDepositReceived.ts
```diff
@@ -4,16 +4,20 @@ import logger from '../../utils/logger';
...
-    principalSwapId: u64.nullable().optional(),
-    gasSwapId: u64.nullable().optional(),
+    principalSwapId: u128.nullable().optional(),
+    gasSwapId: u128.nullable().optional(),
   }),
   z.object({
     __kind: z.literal('NoAction'),
   }),
+  z.object({
+    __kind: z.literal('BoostersCredited'),
```
jerryafr, line 18: should the handler do something with this action? right now it seems like handler will just ignore it

GabrielBuragev: Thats coming in a follow up PR.

jerryafr (approved):

packages/swap/src/event-handlers/v120/networkDepositReceived.ts
GabrielBuragev: The commit message might've been confusing. Will change that.

PR #754, opened 2024-05-22
feat(swap): handle, store and expose swap boost details [WEB-811]

GabrielBuragev:
This handles the `DepositBoosted` & `InsufficientBoostLiquidity` events by storing the details in the database & exposing them via the `/swap` endpoint.
~These details will be used in a follow up PR to expose the boost status via the `/swap` endpoint.~


packages/swap/src/routes/swap.ts
```diff
@@ -277,6 +280,14 @@ router.get(
...
       failedBlockIndex: failedSwap?.failedBlockIndex ?? undefined,
       depositChannelAffiliateBrokers: affiliateBrokers,
+      depositChannelMaxBoostFeeBps: swapDepositChannel?.maxBoostFeeBps,
+      failedBoosts:
+        swapDepositChannel?.failedBoosts.map((failedBoost) => ({
+          failedAtTimestamp: failedBoost.failedAtTimestamp,
+          failedAtBlockIndex: failedBoost.failedAtBlockIndex,
+          amount: failedBoost.amount.toFixed(),
+        })) ?? [],
+      effectiveBoostFeeBps: swap?.effectiveBoostFeeBps,
```
niklasnatter, line 290: think i would not expose this for now if we dont need it

GabrielBuragev: We have it in the `fees` array, so it can be removed.

packages/swap/src/routes/swap.ts
```diff
@@ -277,6 +280,14 @@ router.get(
       failedAt: failedSwap?.failedAt,
       failedBlockIndex: failedSwap?.failedBlockIndex ?? undefined,
       depositChannelAffiliateBrokers: affiliateBrokers,
+      depositChannelMaxBoostFeeBps: swapDepositChannel?.maxBoostFeeBps,
+      failedBoosts:
+        swapDepositChannel?.failedBoosts.map((failedBoost) => ({
+          failedAtTimestamp: failedBoost.failedAtTimestamp,
+          failedAtBlockIndex: failedBoost.failedAtBlockIndex,
+          amount: failedBoost.amount.toFixed(),
+        })) ?? [],
```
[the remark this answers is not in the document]
niklasnatter: this only happens if the liquidity is insufficient, right? would `skippedBoots` be a better term?

packages/swap/src/event-handlers/v140/depositBoosted.ts
```diff
@@ -0,0 +1,44 @@
...
+  // prewitnessedDepositId: u128,
+});
+
+// DepositBoosted event is emitted instead of NetworkDepositReceived event in v140 due to boost
+// We need to update the depositAmount and store the ingress fee just like we do in the NetworkDepositReceived event
+export const depositBoosted = async ({ prisma, event }: EventHandlerArgs) => {
+  const { asset, boostFee, action, ingressFee, amounts } = depositBoostedSchema.parse(event.args);
+
+  if (action.__kind === 'Swap') {
+    const depositAmount = amounts.reduce((acc, [_, amount]) => acc + amount, BigInt(0));
```
niklasnatter, line 23: i think the deposit amount should be the amount before the boost fee (and ingress fee), right? 
i suspect that we need to add the boost fee to this to get that value, but i might be mistaken

packages/swap/src/routes/swap.ts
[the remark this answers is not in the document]
GabrielBuragev: This one is a bit tricky. We will need it on the frontend to be able to transition to the right view (insufficient liquidity / boost found). Do you think checking against the length of `failedBoosts` is enough ( + checking the `effectiveBoostFeeBps`) to determine the status? 
How do we handle multiple swaps on the frontend? There are no transitions (screen stays on the swap completed card) after the first swap completed right? 
In that case, its safe to just check against the length since we will show the transitions only once.

packages/swap/src/event-handlers/v140/depositBoosted.ts
GabrielBuragev: From what i saw in the events, the summed `amounts` array of tuples will hold the initial deposit amount.

niklasnatter: okay cool, nevermind then!

packages/swap/src/routes/swap.ts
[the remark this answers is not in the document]
GabrielBuragev: > this only happens if the liquidity is insufficient, right? would `skippedBoots` be a better term?

Yes, if there was not enough liquidity in the pools, there will be a failedBoost entry. At that point, there is still no swap as the deposit is pending finalization (normal swap flow).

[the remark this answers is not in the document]
niklasnatter: we dont handle multiple swaps in the swapping frontend at all 🙂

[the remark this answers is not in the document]
niklasnatter: so i think for the "not enough liquidity" screen it would be enough to check the length like you said

packages/swap/src/routes/swap.ts
```diff
@@ -277,6 +290,9 @@ router.get(
       failedAt: failedSwap?.failedAt,
       failedBlockIndex: failedSwap?.failedBlockIndex ?? undefined,
       depositChannelAffiliateBrokers: affiliateBrokers,
+      depositChannelMaxBoostFeeBps: swapDepositChannel?.maxBoostFeeBps,
+      effectiveBoostFeeBps: swap?.effectiveBoostFeeBps,
+      boosted,
```
niklasnatter, line 295: what do you think about setting the `effectiveBoostFeeBps` to 0 instead of exposing another property?

niklasnatter: so `effectiveBoostFeeBps` would be undefined until the deposit is prewitnessed. after that, it will be either the effective boost fee or 0 if the deposit was not boosted

GabrielBuragev: Yeah, i guess that would simplify things. I actually exposed the `boosted` property for convenience for integrators. Its not necessary though

niklasnatter (approved):
🥳

packages/swap/src/event-handlers/v140/depositBoosted.ts
acdibble, line 23:
```suggestion
    const depositAmount = amounts.reduce((acc, [, amount]) => acc + amount, BigInt(0));
```

acdibble (approved):
🚀  boost away

packages/swap/src/event-handlers/v140/depositBoosted.ts
```diff
@@ -0,0 +1,44 @@
...
+    const effectiveBoostFeeBps = (boostFee * BigInt(ONE_IN_PIP)) / BigInt(depositAmount.toString());
+
+    await prisma.swap.update({
+      where: { nativeId: action.swapId },
+      data: {
+        effectiveBoostFeeBps: Number(effectiveBoostFeeBps),
+        depositAmount: depositAmount.toString(),
```
jerryafr, line 30: do we need to override depositAmount here?
Can it happen that the initial deposit is different from the amount aggregate here?

jerryafr: or we don't get DepositReceived for boost swaps at all? (I'm forgetting how it was)

jerryafr (approved):

packages/swap/src/event-handlers/v140/depositBoosted.ts
niklasnatter: i think the `DepositBoosted` comes first (on boost) and then the `DepositReceived` once the safety margin is met (and boosters are repaid)

GabrielBuragev: > or we don't get DepositReceived for boost swaps at all? (I'm forgetting how it was)

Thats true, `DepositReceived` never comes in case of a boosted swap. Thats why i left the small comment above this handler to explain this. We basically need to do the same thing here as in the `DepositReceived` event, that is, update the `depositAmount` and insert the `ingressFee`.

GabrielBuragev: > i think the `DepositBoosted` comes first (on boost) and then the `DepositReceived` once the safety margin is met (and boosters are repaid)

`DepositReceived` was basically renamed to `DepositFinalised` and we already have a proxying handler for that.

Additionally, the `DepositFinalised` now has a different action for boosted swaps and normal swaps. For boosted swaps, the action will be `BoostersCredited`, which we ignore in the `depositReceived` handler. For normal swaps, the action is `Swap` which we handle in the `depositReceived` handler.
Due to that, we need to handle the same logic in the `depositReceived` handler regarding updating the `depositAmount` and `ingressFee`.

PR #772, opened 2024-05-30
fix(swap): handle network fee correctly [WEB-1028]

acdibble:


jerryafr (approved):

niklasnatter (approved):

packages/swap/src/event-handlers/swapExecuted.ts
```diff
@@ -54,17 +54,25 @@ export default async function swapExecuted({
...
     swap.destAsset,
-    BigInt(swap.swapInputAmount.toFixed()),
+    swapInputAmount,
     intermediateAmount,
     swapOutput,
   );
 
+  if (swap.srcAsset === 'Usdc') {
+    const networkFee = fees.find((type) => type.type === 'NETWORK')!;
+    swapInputAmount -= BigInt(networkFee.amount);
```
niklasnatter, line 69: do we need to remember to change this again after 1.5?

acdibble: Yeah but I have a ticket already

acdibble: added a comment

PR #773, opened 2024-05-30
chore(swap): include boost event data in swap status

GabrielBuragev:


packages/swap/src/routes/swap.ts
```diff
@@ -295,6 +295,8 @@ router.get(
       depositChannelAffiliateBrokers: affiliateBrokers,
       depositChannelMaxBoostFeeBps: channel?.maxBoostFeeBps,
       effectiveBoostFeeBps,
+      swapBoostedAt: swap?.swapBoostedAt?.valueOf(),
```
niklasnatter, line 298: maybe `depositBoostedAt` would be more correct?

niklasnatter: now that i think of it, wouldnt this be the same as `depositReceivedAt`

niklasnatter: and should we have a `boostSkippedAt` too?

GabrielBuragev: > now that i think of it, wouldnt this be the same as `depositReceivedAt`

ahh yeah, now i remember why i didn't include it int he first place 🙈

packages/swap/src/routes/swap.ts
```diff
@@ -295,6 +295,10 @@ router.get(
       depositChannelAffiliateBrokers: affiliateBrokers,
       depositChannelMaxBoostFeeBps: channel?.maxBoostFeeBps,
       effectiveBoostFeeBps,
+      depositBoostedAt: swap?.depositBoostedAt?.valueOf(),
```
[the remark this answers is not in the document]
niklasnatter: also would be nice to add a test case that has the properties

niklasnatter (approved):

packages/swap/src/routes/swap.ts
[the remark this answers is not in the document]
GabrielBuragev: Done

packages/sdk/src/swap/types.ts
```diff
@@ -82,6 +82,8 @@ interface DepositAddressFields extends SwapStatusResponseCommonFields {
   depositChannelAffiliateBrokers?: { account: string; commissionBps: number }[];
   depositChannelMaxBoostFeeBps: number;
   effectiveBoostFeeBps?: number;
+  boostSkippedAt?: number;
+  boostKippedBlockIndex?: string;
```
jerryafr, line 86:
```suggestion
  boostSkippedBlockIndex?: string;
```

PR #776, opened 2024-06-07
fix(swap): return confirmation count only after 2 state chain blocks

jerryafr:


packages/swap/src/ingress-egress-tracking/index.ts
```diff
@@ -49,13 +49,19 @@ export const getPendingDeposit = async (
...
 
-    if (deposits.length === 0 || tracking == null) {
+    const currentHeight = (await prisma.state.findFirstOrThrow()).height;
+    // For boosted swaps, state chain will track the foreign chain block one state chain block before the DepositBoosted event
+    // Because of this, frontend will jump to confirmations page and then, will switch back to checking boost liquidity
+    // To prevent this, we need to wait for 2 state chain blocks before checking the confirmations
+    const isOneBlockPassedTwoBlocks =
+      currentHeight - (tracking?.blockTrackedAtStateChainBlock ?? 0) > 2;
+
+    if (deposits.length === 0 || tracking == null || !isOneBlockPassedTwoBlocks) {
```
niklasnatter, line 59: im afraid now we would jump back every time we track a new block? 😕

packages/swap/prisma/schema.prisma
```diff
@@ -317,11 +317,13 @@ model BoostPool {
...
-  id             Int      @id @default(autoincrement())
-  chain          Chain    @unique
-  height         BigInt   @default(0)
-  blockTrackedAt DateTime @default(now())
-  updatedAt      DateTime @updatedAt @db.Timestamptz(3)
+  id                            Int      @id @default(autoincrement())
+  chain                         Chain    @unique
+  height                        BigInt   @default(0)
+  previousHeight                BigInt   @default(0)
+  blockTrackedAtStateChainBlock Int
```
zoheb391, line 324: i think we can improve this name. What exactly is this ? the statechain block that saw this event?

jerryafr: yes exactly, suggestions are really appreciated.
Also for `isTrackingPassedTwoStateChainBlocks` :-/

zoheb391: how about `eventBlock` or `eventWitnessedBlock` ?

zoheb391: also is there a reason we need previousHeight as a db entry?

packages/swap/src/ingress-egress-tracking/index.ts
```diff
@@ -56,9 +56,18 @@ export const getPendingDeposit = async (
       return mempoolTx && { ...mempoolTx, transactionConfirmations: 0 };
     }
 
+    const currentHeight = (await prisma.state.findFirstOrThrow()).height;
+    // For boosted swaps, state chain will track the foreign chain block one state chain block before the DepositBoosted event
+    // Because of this, frontend will jump to confirmations page and then, will switch back to checking boost liquidity
+    // To prevent this, we need to wait for 1 state chain blocks before checking the confirmations
+    const isTrackingPassedTwoStateChainBlocks =
+      currentHeight - (tracking?.blockTrackedAtStateChainBlock ?? 0) > 0;
+
```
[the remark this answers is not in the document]
niklasnatter: then we could tell the story why this is needed somehow like this:
```
// deposits in an external block will be witnessed (and boosted) one statechain block after chain tracking is updated
// to prevent jumping from "1 confirmation" to "checking boost liquidity", we need to calculate confirmations based on the last witnessed block
const statechainBlocksSinceTracking = currentHeight - (tracking?.blockTrackedAtStateChainBlock ?? 0);
const lastWitnessedBlockHeight = statechainBlocksSinceTracking > 0 ? tracking.height : tracking.previousHeight

...
confirmations = statechainBlocksSinceTracking - deposits[0].deposit_chain_block_height + 1
```

niklasnatter (approved):

packages/swap/prisma/schema.prisma
jerryafr: updated to `eventWitnessedBlock`

zoheb391 (approved):

PR #802, opened 2024-07-01
feat(sdk): expose sdk method for fetching boost liquidity

GabrielBuragev:


jerryafr (approved):

packages/sdk/src/swap/sdk.ts
```diff
@@ -102,6 +111,12 @@ export class SwapSDK {
     return this.supportedAssets;
   }
 
+  private async getBoostPoolsDepth(): Promise<BoostPoolsDepth> {
+    this.boostPoolsDepth ??= await getAllBoostPoolsDepth(this.rpcConfig);
+
+    return this.boostPoolsDepth;
```
acdibble, line 117: if we memoize this the data will get stale quickly
```suggestion
    return getAllBoostPoolsDepth(this.rpcConfig);
```

GabrielBuragev: Done

packages/sdk/src/swap/types.ts
```diff
@@ -211,3 +211,10 @@ export type SwapStatusResponse =
   | DepositAddressStatusResponse
   | VaultSwapStatusResponse
   | FailedVaultSwapStatusResponse;
+
+export type BoostPoolDepth = {
+  tier: number;
```
niklasnatter, line 216: should we call this something like `feeTierBps`?

niklasnatter: i think atm we can still choose 🙂 once we exposed a name somewhere we have to keep it

niklasnatter: could also call it `feeBps` if that is more consistent with the other terms like `effectiveBoostFeeBps`

acdibble (approved):

packages/sdk/src/swap/types.ts
GabrielBuragev: Yes, lets call it `feeTierBps`. I also want to communicate the idea of tiers.

GabrielBuragev: Done

niklasnatter (approved):

PR #809, opened 2024-07-08
feat(cli): Add refundParameters to requestSwapDepositAddress method [WEB-1253]

niklasnatter:


packages/cli/src/commands/cliRequestSwapDepositAddress.ts
```diff
@@ -68,7 +68,14 @@ export default async function cliRequestSwapDepositAddress(
       destAddress: args.destAddress,
       ccmMetadata,
       commissionBps: args.commission ?? 0,
-      affiliates: [],
+      affiliates: [
```
niklasnatter, line 71: TODO: revert

acdibble (approved):

jerryafr (approved):

packages/shared/src/schemas.ts
```diff
@@ -78,6 +86,7 @@ export const openSwapDepositChannelSchema = z
     ccmMetadata: ccmMetadataSchema.optional(),
     maxBoostFeeBps: z.number().optional(),
     srcAddress: z.string().optional(),
+    refundParameters: refundParameters.optional(),
```
zoheb391, line 89: is `refundParameters` the best name for this? its basically the parameters for fillOrKill right? so i would call it `fillOrKillOptions` so it sounds like the feature that it is.

PR #811, opened 2024-07-11
feat(swap): Store deposit transaction ref on swap [WEB-1189]

niklasnatter:


packages/swap/src/event-handlers/networkBroadcastSuccess.ts
```diff
@@ -14,7 +14,7 @@ export const networkBroadcastSuccessArgs = (chain: Chain) =>
           .transform((v) => `${v.blockNumber}-${v.extrinsicIndex}`),
         hexString,
       ])
-      .transform((v) => (chain === 'Bitcoin' ? reverseBytes(v) : v))
+      .transform((v) => (chain === 'Bitcoin' ? reverseBytes(v.substring(2)) : v))
```
niklasnatter, line 17: we exposed the bitcoin tx hash with a leading `0x` till now (bitcoin explorers do not support that when i tested) 😕 will add a migration for the existing broadcasts

packages/sdk/src/swap/types.ts
```diff
@@ -129,7 +129,8 @@ type SwapState =
   | {
       state: 'AWAITING_DEPOSIT';
       depositAmount: string | undefined;
-      depositTransactionHash: string | undefined;
+      depositTransactionHash: string | undefined; // DEPRECATED(1.5): use depositTransactionRef instead
```
niklasnatter, line 132: unfortuante existing naming bc the polkadot thing is not a hash (and we call it `ref` for broadcasts). will expose both names for now

acdibble (approved):

packages/swap/prisma/migrations/20240711155734_migrate_bitcoin_transaction_ref/migration.sql
```diff
@@ -0,0 +1,2 @@
+UPDATE "public"."Broadcast" SET "transactionRef" = SUBSTRING("transactionRef", 3) WHERE "chain" = 'Bitcoin' AND "transactionRef" IS NOT NULL;
```
jerryafr, line 1: is 3 correct here? shouldn't this be substring( ..., 2) ?

jerryafr (approved):

packages/swap/prisma/migrations/20240711155734_migrate_bitcoin_transaction_ref/migration.sql
niklasnatter: surprisingly yes - if you pass 1 the first character will still be included 🙈

packages/swap/src/event-handlers/v120/networkDepositReceived.ts
```diff
@@ -1,19 +1,58 @@
...
+  evmDepositDetails,
+  polkadotDepositDetails,
+]);
+
 export const depositReceivedArgs = z.object({
   amount: u128,
   asset: internalAssetEnum,
   ingressFee: u128,
   action: actionSchema,
+  blockHeight: u128.optional(),
```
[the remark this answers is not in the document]
niklasnatter: yes that makes sense!

PR #815, opened 2024-07-17
chore(shared): Move commissionBps and affiliates into correct param [WEB-1253]

niklasnatter:


packages/shared/src/broker.ts
```diff
@@ -82,11 +82,7 @@ export type DepositChannelResponse = ReturnType<typeof validateResponse>;
 
 export async function requestSwapDepositAddress(
```
niklasnatter, line 83: this method is not exposed from the sdk package, so its fine to change it

acdibble (approved):

jerryafr (approved):

PR #821, opened 2024-07-22
chore(sdk): Add SDK version header to allow for backwards compatibility code on the server

niklasnatter:


packages/sdk/src/swap/services/ApiService.ts
```diff
@@ -24,12 +25,14 @@ export const getQuote: BackendQuery<
     }),
   };
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const queryParams = new URLSearchParams(params as Record<string, any>);
-
-  const url = new URL(`/quote?${queryParams.toString()}`, baseUrl).toString();
```
niklasnatter, line 30: found out that axios will do this for us while looking for the headers option

jerryafr (approved):

packages/sdk/src/swap/consts.ts
```diff
@@ -1,4 +1,7 @@
 import { ChainflipNetworks } from '@/shared/enums';
+import { version } from '../../package.json';
+
+export const CF_SDK_VERSION_HEADERS = { 'Cf-Sdk-Version': version };
```
acdibble, line 4:
```suggestion
export const CF_SDK_VERSION_HEADERS = { 'X-Cf-Sdk-Version': version };
```

packages/sdk/src/swap/consts.ts
```diff
@@ -1,4 +1,7 @@
 import { ChainflipNetworks } from '@/shared/enums';
+import { version } from '../../package.json';
```
acdibble, line 2: does this get inlined or does it have the entire package.json in the built output

packages/sdk/src/swap/services/ApiService.ts
```diff
@@ -24,12 +25,12 @@ export const getQuote: BackendQuery<
...
-
-  const { data } = await axios.get<QuoteQueryResponse>(url, { signal });
+  const { data } = await axios.get<QuoteQueryResponse>('/quote', {
```
acdibble, line 28: we can use `axios.create` to create an instance that automatically sets the headers and then share it everywhere

packages/sdk/src/swap/consts.ts
niklasnatter: i think its inlined - found this in the build: 
Screenshot 2024-07-23 at 10 03 11

packages/sdk/src/swap/services/ApiService.ts
niklasnatter: unfortunately we cant pass the `baseUrl` via `axios.create` bc we pass it to the methods, so the instance would only contain the headers. no strong opinion about it

packages/sdk/src/swap/consts.ts
niklasnatter: i asked chatgpt how to name the header yesterday lol, it told me that the `X-` prefix is not a recommendation anymore. but no strong opinion, im fine with adding it if you want

packages/sdk/src/swap/services/ApiService.ts
acdibble: you can pass a base url: https://axios-http.com/docs/config_defaults

acdibble: i c

acdibble (approved):

PR #830, opened 2024-07-31
feat: solana

acdibble:


packages/shared/src/__tests__/schema.test.ts
```diff
@@ -48,19 +51,21 @@ describe('postSwapSchema', () => {
...
   });
+
   it('handles missing ccm params', () => {
     expect(
       openSwapDepositChannelSchema.safeParse({
         ...swapBody,
         ccmParams: {
           message: '0xdeadc0de',
-          cfParameters: 'string',
+          cfParameters: '0xcafebabe',
```
GabrielBuragev, line 63: `0xcafebabe` is the new `0xdeadc0de`

packages/shared/src/enums.ts
```diff
@@ -125,6 +141,12 @@ export const chainConstants = {
     contractId: 4,
     blockTimeSeconds: 0.26,
   },
+  [Chains.Solana]: {
+    assets: [Assets.SOL, Assets.USDC],
+    gasAsset: Assets.SOL,
+    contractId: 5,
+    blockTimeSeconds: (400 + 800) / 2 / 1000,
```
GabrielBuragev, line 148: how is this calculated? Do you have some reference by any chhance?

packages/shared/src/validation/addressValidation.ts
```diff
@@ -26,36 +27,50 @@ export const validateBitcoinTestnetAddress: AddressValidator = (address: string)
...
   bitcoin.isValidAddressForNetwork(address, 'regtest');
 
+export const validateSolanaAddress: AddressValidator = (address: string) => {
+  try {
+    base58.decode(address);
+    return true;
+  } catch {
+    return false;
+  }
+};
```
GabrielBuragev, line 37: can we move this to the `@chainflip/solana` package?
We will end up with 2 different validation functions (one more strict than the other), but i think they will both be useful and we might need the one that is less-strict (this one) on the frontends as well.

GabrielBuragev (approved):

zoheb391 (approved):

packages/swap/src/polkadot/api.ts
```diff
@@ -10,6 +10,7 @@ export const boostSafeModeSchema = z.object({
   ingressEgressBitcoin: boostFlagsSchema,
   ingressEgressPolkadot: boostFlagsSchema,
   ingressEgressArbitrum: boostFlagsSchema,
+  ingressEgressSolana: boostFlagsSchema.optional(),
```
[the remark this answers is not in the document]
GabrielBuragev: to preserve backwards compatibility i think

packages/shared/src/enums.ts
acdibble: No I just read that it takes between 400ms and 800ms for a block, so i took the average

GabrielBuragev:
There is a problem with this sdk version that causes the frontends to crash with the following error:

Screenshot 2024-08-23 at 16 45 58

GabrielBuragev:
Problem should be fixed now.

PR #840, opened 2024-08-06
chore: change getBoostLiquidity fn signature

GabrielBuragev:


niklasnatter (approved):
seems like this is not tested - can we add a test for it?

GabrielBuragev:
> seems like this is not tested - can we add a test for it?

It actually is tested, forgot to adjust the tests :D

zoheb391 (approved):

jerryafr (approved):

PR #843, opened 2024-08-06
chore: Accept minPrice instead of minPriceX128 in requestDepositAddress [WEB-1253]

niklasnatter:


GabrielBuragev (approved):

packages/swap/src/routes/swap.ts
```diff
@@ -327,13 +326,11 @@ router.get(
...
-              .shiftedBy(
-                assetConstants[channel.srcAsset].decimals -
-                  assetConstants[channel.destAsset].decimals,
-              )
-              .toFixed(),
+            minPrice: getPriceFromPriceX128(
+              channel.fokMinPriceX128.toFixed(),
+              channel.srcAsset,
+              channel.destAsset,
+            ),
```
jerryafr, line 333: 🙌

jerryafr (approved):

packages/sdk/src/swap/__tests__/sdk.test.ts
```diff
@@ -326,6 +326,7 @@ describe(SwapSDK, () => {
           retryDurationBlocks: 500,
           refundAddress: '0xa56A6be23b6Cf39D9448FF6e897C29c41c8fbDFF',
           minPrice: '10000000000000',
+          minPriceX128: '34028236692093846346337460743176821145600000000000000000000000',
```
jerryafr, line 329: should we send both to the rpc?

niklasnatter: its not sent to the chainflip rpc, it is the `trpc` handler. i think it should not hurt and it was annoying to remove, so i left it 😅

PR #845, opened 2024-08-06
feat(swap,sdk): [WEB-1311] add swapRescheduled info to getStatus

jerryafr:


zoheb391 (approved):

packages/sdk/src/swap/types.ts
```diff
@@ -82,6 +82,9 @@ interface SwapStatusResponseCommonFields extends ChainsAndAssets {
   estimatedDefaultDurationSeconds: number | undefined;
   srcChainRequiredBlockConfirmations: number | undefined;
   depositTransactionRef: string | undefined;
+  latestSwapScheduledAt: number | undefined;
+  latestSwapScheduledBlockIndex: string | undefined;
+  retryCount: number | undefined;
```
niklasnatter, line 87: i think we arent sure yet if we want to expose this or if it should be an implementation detail, so maybe remove it from the route and type for now? think it makes sense to keep it in the db

niklasnatter (approved):
otherwise looks good to me

PR #847, opened 2024-08-07
feat(sdk,swap): [WEB-1283] handle refundEgressScheduled and add status

jerryafr:


acdibble (approved):

packages/swap/src/event-handlers/refundEgressScheduled.ts
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  await prisma.swap.update({
+    where: { nativeId: swapId },
+    data: {
+      refundEgress: { connect: { id: egress.id } },
+      fees: {
+        create: {
+          type: 'EGRESS',
+          asset: swap.destAsset,
```
niklasnatter, line 47: i think this will be the src asset in this case

niklasnatter (approved):
looks good otherwise!

packages/swap/src/event-handlers/refundEgressScheduled.ts
jerryafr: yes true, good catch!

PR #851, opened 2024-08-08
fix(swap): refund address parser should use foreignChainAddress format

jerryafr:


packages/swap/src/event-handlers/swapDepositAddressReady.ts
```diff
@@ -25,7 +25,7 @@ const swapDepositAddressReadyArgs = z.object({
   refundParameters: z
     .object({
       minPrice: u128,
-      refundAddress: encodedAddress,
+      refundAddress: foreignChainAddress('backspin'),
```
acdibble, line 28: env var

jerryafr: 😨

niklasnatter (approved):

jerryafr:
Good thing I haven't merged it yet.
For visibility: we don't need this anymore as protocol changed the shape so refund address will be encodedAddress as well

PR #852, opened 2024-08-08
feat(swap,sdk): [WEB-1284] add REFUNDED status

jerryafr:


packages/swap/src/routes/swap.ts
```diff
@@ -221,6 +222,8 @@ router.get(
       if (swap.refundEgress.broadcast?.abortedAt) {
         state = State.Failed;
         failureMode = Failure.RefundBroadcastAborted;
+      } else if (swap.refundEgress.broadcast?.succeededAt) {
+        state = State.Refunded;
```
acdibble, line 226: do we have a refund tx hash?

acdibble (approved):

packages/sdk/src/swap/types.ts
```diff
@@ -209,6 +209,12 @@ type SwapState =
       broadcastSucceededBlockIndex: string;
       broadcastTransactionRef: string;
     } & BroadcastRequested)
+  | ({
+      state: 'REFUNDED';
+      broadcastSucceededAt: number;
+      broadcastSucceededBlockIndex: string;
+      broadcastTransactionRef: string;
+    } & BroadcastRequested)
```
zoheb391, line 217: i dont think you can use `BroadcastRequested` because it has `swapExecutedAt`

packages/swap/src/routes/swap.ts
niklasnatter: should be able to use the one from the broadcast

packages/sdk/src/swap/types.ts
jerryafr: yeah, that's true
updated

packages/swap/src/routes/swap.ts
jerryafr: yeah it's a normal broadcast procedure, so we should have

niklasnatter (approved):

PR #857, opened 2024-08-12
feat(swap): [WEB-1285] handle RefundEgressIgnored and expose via FAILED status

jerryafr:


packages/swap/src/routes/swap.ts
```diff
@@ -43,6 +43,7 @@ export enum State {
 export enum Failure {
   IngressIgnored = 'INGRESS_IGNORED',
   EgressIgnored = 'EGRESS_IGNORED',
+  RefundEgressIgnored = 'REFUND_EGRESS_IGNORED',
```
niklasnatter, line 46: think you need to update the type exposed from the sdk as well

niklasnatter (approved):

acdibble (approved):

GabrielBuragev (approved):

PR #860, opened 2024-08-13
fix(swap): always query deposit channel

acdibble:
Deposit channel info present here:
https://chainflip-swap-perseverance.chainflip.io/swaps/16702-Ethereum-15
but absent here:
https://chainflip-swap-perseverance.chainflip.io/swaps/10

### DIFF
--- a/packages/swap/src/routes/swap.ts
+++ b/packages/swap/src/routes/swap.ts
@@ -6,18 +6,7 @@ import { openSwapDepositChannelSchema } from '@/shared/schemas';
 import { screamingSnakeToPascalCase, toUpperCase } from '@/shared/strings';
 import { getRequiredBlockConfirmations } from '@/swap/utils/rpc';
 import { asyncHandler, maintenanceMode } from './common';
-import prisma, {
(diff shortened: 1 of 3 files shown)


packages/swap/src/routes/swap.ts
```diff
@@ -160,14 +136,18 @@ router.get(
...
       });
       if (!swap) {
         failedSwap = await prisma.failedSwap.findFirst({
           where: { depositTransactionRef: id },
+          include: { swapDepositChannel: { include: depositChannelInclude } },
         });
       }
     }
 
+    swapDepositChannel ??= swap?.swapDepositChannel ?? failedSwap?.swapDepositChannel;
```
niklasnatter, line 149: would be nice to check this in one of the tests

niklasnatter (approved):

packages/swap/src/routes/swap.ts
acdibble: just added it

jerryafr (approved):

PR #892, opened 2024-08-30
refactor(swap): add usd value to log

acdibble:


packages/swap/src/utils/getPoolQuote.ts
```diff
@@ -110,6 +110,7 @@ export default async function getPoolQuote({
     egressAmount: outputAmount.toString(),
     includedFees,
     lowLiquidityWarning,
+    inputUsdValue: boostFeeBps ? inputUsdValue : undefined,
```
GabrielBuragev, line 113: why are we checking for `boostFeeBps` here?

acdibble: oh it's the wrong way around. i don't want to include it on boosted quotes

zoheb391 (approved):

GabrielBuragev (approved):

PR #902, opened 2024-09-02
fix: handle boostable lp channels

GabrielBuragev:


packages/swap/src/event-handlers/v140/insufficientBoostLiquidity.ts
```diff
@@ -12,20 +13,38 @@ export const insufficientBoostLiquiditySchema = z.object({
 export const insufficientBoostLiquidity = async ({ prisma, event, block }: EventHandlerArgs) => {
   const { channelId, asset, amountAttempted } = insufficientBoostLiquiditySchema.parse(event.args);
 
-  const channel = await prisma.swapDepositChannel.findFirstOrThrow({
-    where: { channelId, srcAsset: asset },
+  const depositChannel = await prisma.depositChannel.findFirst({
+    where: { channelId, srcChain: assetConstants[asset].chain },
```
acdibble, line 17:
```suggestion
    where: { channelId, srcChain: assetConstants[asset].chain, isSwapping: true },
```

packages/swap/src/event-handlers/v140/insufficientBoostLiquidity.ts
```diff
@@ -12,20 +13,38 @@ export const insufficientBoostLiquiditySchema = z.object({
...
-      swapDepositChannel: {
-        connect: {
-          id: channel.id,
+  if (!depositChannel) {
+    throw new Error(
+      `InsufficientBoostLiquidity: Deposit channel not found for asset ${asset} and channelId ${channelId}`,
+    );
+  }
+
+  if (depositChannel.isSwapping) {
```
acdibble, line 29:
```suggestion
```

acdibble (approved):
slight simplification

packages/swap/src/event-handlers/v140/insufficientBoostLiquidity.ts
GabrielBuragev: I wanted to still keep the assertion and check for an existing deposit channel disregarding the type of the channel

PR #903, opened 2024-09-03
feat(swap): add v2 status endpoint (DCA)

zoheb391:
closes WEB-1363

before:
```ts
export enum State {
  Failed = 'FAILED',
  Complete = 'COMPLETE',
  BroadcastAborted = 'BROADCAST_ABORTED', // TODO: move to Failed state
  Broadcasted = 'BROADCASTED',
  BroadcastRequested = 'BROADCAST_REQUESTED',
  EgressScheduled = 'EGRESS_SCHEDULED',
  SwapExecuted = 'SWAP_EXECUTED',
  DepositReceived = 'DEPOSIT_RECEIVED',
  AwaitingDeposit = 'AWAITING_DEPOSIT',
}
```
now:
```ts
export enum StateV2 {
  Failed = 'FAILED',
  Complete = 'COMPLETE',
  Sent = 'SENT',
  Sending = 'SENDING',
  Swapping = 'SWAPPING',
  Receiving = 'RECEIVING',
}
```

Some big differences:
- each main field can have its owns failure object (swap, deposit, refund)
- no more `broadcast.requestedAt` since egress.scheduledAt serves the same purpose.
- broadcast aborted is no longer its own state. It comes under `FAILED`
- rename `egress` and `broadcast` to `schedule` and `sent`
- aggregation of fees which now includes broker
- tried to extract as much as possible to `./utils` to make this file more legible.


packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,408 @@
...
+      pendingDeposit?.transactionHash ??
+      failedSwap?.depositTransactionRef ??
+      undefined;
+
+    const sortedSwaps = swapRequest?.swaps.sort((a, b) => {
+      if (a.swapExecutedAt && b.swapExecutedAt) {
+        return a.swapExecutedAt.valueOf() - b.swapExecutedAt.valueOf() ? 1 : -1;
+      }
+      return a.swapExecutedAt && !b.swapExecutedAt ? 1 : 0;
+    });
```
zoheb391, line 263: new

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,408 @@
...
+          },
+          {
+            totalAmountSwapped: new Prisma.Decimal(0),
+            totalChunksExecuted: 0,
+            currentChunk: sortedSwaps[0],
+            lastExecutedChunk: null as null | ReturnType<typeof getSwapFields>,
+            isDcaSwap:
+              swapDepositChannel?.chunkIntervalBlocks && swapDepositChannel.chunkIntervalBlocks > 1,
+          },
+        ),
```
zoheb391, line 328: new

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,408 @@
...
+            where: { id: swapRequest.ignoredEgresses.at(0)!.stateChainErrorId },
+          }),
+        ]);
+        error = {
+          name: stateChainError.name,
+          message: stateChainError.docs,
+        };
+      } else if (hasAbortedBroadcasts) {
+        failureMode = Failure.BroadcastAborted;
+      }
```
zoheb391, line 195: few changes here

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,408 @@
...
+    } else if (egress?.broadcast) {
+      const pendingBroadcast = await getPendingBroadcast(egress.broadcast);
+      if (pendingBroadcast) {
+        state = StateV2.Broadcasted;
+        egressTrackerTxRef = pendingBroadcast.tx_ref;
+      }
+    } else if (egress) {
+      state = StateV2.EgressScheduled;
+    } else if (swapRequest?.swaps.some((s) => s.swapExecutedAt)) {
+      state = StateV2.Swapping;
```
zoheb391, line 207: new

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,408 @@
...
+} from '../../utils/swap';
+import { asyncHandler } from '../common';
+import { Failure } from '../swap';
+
+const router = express.Router();
+
+export enum StateV2 {
+  Failed = 'FAILED',
+  Complete = 'COMPLETE',
+  Broadcasted = 'BROADCASTED',
```
zoheb391, line 27: removed broadcast aborted

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,408 @@
...
+      }
+    } else if (egress) {
+      state = StateV2.EgressScheduled;
+    } else if (swapRequest?.swaps.some((s) => s.swapExecutedAt)) {
+      state = StateV2.Swapping;
+    } else if (swapRequest?.depositReceivedAt) {
+      state = StateV2.DepositReceived;
+    } else {
+      state = StateV2.AwaitingDeposit;
+    }
```
zoheb391, line 212: few changes in this ssection

packages/swap/src/routes/v2/swap.ts
zoheb391, line 328: new

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,408 @@
...
+              acc.totalAmountSwapped = acc.totalAmountSwapped.plus(curr.swapOutputAmount ?? 0);
+              if (curr.swapScheduledAt && curr.swapExecutedAt) {
+                acc.lastExecutedChunk = getSwapFields(curr);
+                acc.totalChunksExecuted += 1;
+              }
+            }
+            return acc;
+          },
+          {
+            totalAmountSwapped: new Prisma.Decimal(0),
```
acdibble, line 321: probably need to call `toFixed` on this to prevent exponential notation when `toString` is implicitly called

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,408 @@
...
+          txRef: egress?.broadcast?.transactionRef ?? egressTrackerTxRef,
+        },
+        srcChainRequiredBlockConfirmations:
+          (internalSrcAsset && (await getRequiredBlockConfirmations(internalSrcAsset))) ??
+          undefined,
+        estimatedDefaultDurationSeconds:
+          srcAsset && destAsset && (await estimateSwapDuration({ srcAsset, destAsset })),
+      },
+      refund: {
+        egress: {
```
acdibble, line 358: would probably flatten this a bit and remove `egress` and `broadcast`. I think we exposed too many internal implementation details in the original one. Instead we can do something more like:
```ts
refund?: {
  scheduledAt: Date;
  scheduledAtBlockIndex: string;
  failedAt?: Date;
  failedAtBlockIndex?: string;
  failureReason?: string;
  completedAt?: Date;
  completedAtBlockIndex?: string;
}
```

if the broadcast is aborted or the egress is ignored, we can just use the `failedAt` field and included failure info to describe the specific mode of failure.

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,408 @@
...
+        receivedAt: swapRequest?.depositReceivedAt?.valueOf(),
+        receivedBlockIndex: swapRequest?.depositReceivedBlockIndex ?? undefined,
+      },
+      swap: {
+        ...sortedSwaps?.reduce(
+          (acc, curr) => {
+            if (isEgressableSwap(curr)) {
+              acc.totalAmountSwapped = acc.totalAmountSwapped.plus(curr.swapOutputAmount ?? 0);
+              if (curr.swapScheduledAt && curr.swapExecutedAt) {
+                acc.lastExecutedChunk = getSwapFields(curr);
```
acdibble, line 314: would probably do this and then transform only after reduce has finished
```suggestion
                acc.lastExecutedChunk = curr;
```

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,408 @@
...
+      failedSwap?.depositTransactionRef ??
+      undefined;
+
+    const sortedSwaps = swapRequest?.swaps.sort((a, b) => {
+      if (a.swapExecutedAt && b.swapExecutedAt) {
+        return a.swapExecutedAt.valueOf() - b.swapExecutedAt.valueOf() ? 1 : -1;
+      }
+      return a.swapExecutedAt && !b.swapExecutedAt ? 1 : 0;
+    });
+    // const fees = (swapRequest?.fees ?? []).concat(swap?.fees ?? []);
```
acdibble, line 264: maybe we can add the swap id or the swap request id to the fees to make it clear where they came from but allow us to put them all in one array still

packages/swap/src/routes/v2/swap.ts
acdibble: i think egress scheduled can be removed because it always happens in the same block as broadcast requested afaik.

we should reduce this further into maybe more high level waiting, receiving, swapping, sending, sent (ingress/egress tracker), complete. and if we have more information available we can put it into some other substatus i think. probably better to just have broad strokes here

zoheb391: yeah we removed `broadcast_requested` because its the same as `egress_scheduled` basically. But we keep broadcasted as that expects the tx in the mempool

zoheb391: i'm open to renaming some of these statuses to simpler language. will try my hand at it first.

packages/swap/src/routes/v2/swap.ts
zoheb391: i'll wait for the broker fee refactor to handle fees. I think that might change how we want to structure it.

zoheb391: didn't realize it was done. will do it here

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,438 @@
...
+    }),
+    ...(broadcast && {
+      sentAt: broadcast?.succeededAt?.valueOf(),
+      sentAtBlockIndex: broadcast?.succeededBlockIndex ?? undefined,
+      sentTxRef: broadcast?.transactionRef ?? egressTrackerTxRef,
+    }),
+    ...(failedAt && failedAtBlockIndex && { failedAt, failedAtBlockIndex }),
+    ...(ignoredEgress && { ignoredAmount: ignoredEgress.amount?.toFixed() }),
+  };
+};
```
zoheb391, line 111: renamed some of these fields to make it more human readable. I've removed `broadcast.requestedAt` since its basically the same as `egressScheduledAt`

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,438 @@
...
+          swapEgressTrackerTxRef = pendingSwapBroadcast.tx_ref;
+        }
+      }
+      if (refundEgress?.broadcast) {
+        const pendingRefundBroadcast = await getPendingBroadcast(refundEgress.broadcast);
+        if (pendingRefundBroadcast) {
+          state = StateV2.Broadcasted;
+          refundEgressTrackerTxRef = pendingRefundBroadcast.tx_ref;
+        }
+      }
```
zoheb391, line 261: we have two trackerTxRefs now

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,438 @@
...
+        ]);
+        error = {
+          name: stateChainError.name,
+          message: stateChainError.docs,
+        };
+      } else if (hasAbortedBroadcasts) {
+        const message = swapEgress?.broadcast?.abortedAt
+          ? 'The swap broadcast was aborted'
+          : 'The refund broadcast was aborted';
+        failureMode = Failure.BroadcastAborted;
```
zoheb391, line 239: broadcastAborted is now a failed status

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,438 @@
...
+          'SWAP',
+          refundEgressTrackerTxRef,
+          failedSwap,
+        ),
+      },
+      ccm: {
+        ...ccmParams,
+        receivedBlockIndex: swapRequest?.ccmDepositReceivedBlockIndex ?? undefined,
+      },
+      boost: {
```
acdibble, line 422: i'd have these top level properties absent/null i they don't apply to a given swap/channel

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,323 @@
...
+  estimateSwapDuration,
+  isEgressableSwap,
+} from '../../utils/swap';
+import { asyncHandler } from '../common';
+
+const router = express.Router();
+
+export enum StateV2 {
+  Failed = 'FAILED',
+  Complete = 'COMPLETE',
```
jerryafr, line 32: just a nitpick, maybe past tense would be better here as well.
```suggestion
  Completed = 'COMPLETED',
```

packages/swap/src/routes/swap.ts
```diff
@@ -177,20 +162,21 @@ router.get(
           name: failedSwap.reason,
           message: failedSwapMessage[failedSwap.reason],
         };
-      } else if (swapRequest?.ignoredEgress) {
-        switch (swapRequest.ignoredEgress.type) {
+      } else if (hasIgnoredEgresses) {
+        const ignored = swapRequest?.ignoredEgresses.at(0);
+        switch (ignored!.type) {
           case 'REFUND':
             failureMode = Failure.RefundEgressIgnored;
```
jerryafr, line 169: I feel like it's better to consider a REFUNDED as a Completed swap.
We can keep Failed only for the cases when something went wrong and the money was not received by the sender or the receiver.
What do you think?

zoheb391: thats only the case when the refund egress is ignored.. Refunded is still a completed swap

jerryafr: ah correct. you're right
Thanks!

jerryafr (approved):

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,323 @@
...
+          ? {
+              numberOfChunks: swapDepositChannel?.numberOfChunks,
+              chunkIntervalBlocks: swapDepositChannel?.chunkIntervalBlocks,
+            }
+          : undefined,
+      },
+      deposit: {
+        amount:
+          readField(swapRequest, failedSwap, 'depositAmount')?.toFixed() ?? pendingDeposit?.amount,
+        transactionRef: depositTransactionRef,
```
niklasnatter, line 275: its called `transactionRef` but `sentTxRef` below - we should adjust one of them to make it consistent

niklasnatter: if we split up `swap` into `swap`+`egress` we could call it `transactionRef` on the `egress` object (and the `refund` object)

packages/swap/src/routes/__tests__/v2/swap.test.ts
```diff
@@ -0,0 +1,1817 @@
...
+          brokerCommissionBps: 0,
+          depositAddress: '0x6aa69332b63bb5b1d7ca5355387edd5624e181f2',
+          expiryBlock: '265',
+          estimatedExpiryTime: 1699527060000,
+          isExpired: false,
+          openedThroughBackend: false,
+        },
+        deposit: {},
+        swap: {
+          srcChainRequiredBlockConfirmations: 2,
```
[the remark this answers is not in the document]
niklasnatter: oky i think `depositChannel` would be even better

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,323 @@
...
+      state,
+      swapId: swapRequest?.nativeId.toString(),
+      ...(internalSrcAsset && getAssetAndChain(internalSrcAsset, 'src')),
+      ...(internalDestAsset && getAssetAndChain(internalDestAsset, 'dest')),
+      destAddress: readField(swapRequest, swapDepositChannel, failedSwap, 'destAddress'),
+      depositChannel: {
+        createdAt: swapDepositChannel?.createdAt.valueOf(),
+        brokerCommissionBps: swapDepositChannel?.brokerCommissionBps,
+        depositAddress: swapDepositChannel?.depositAddress,
+        expiryBlock: swapDepositChannel?.srcChainExpiryBlock?.toString(),
```
niklasnatter, line 248: would keep the `srcChain` prefix because we also have `retryDurationBlocks` which is a different kind of block 😅

niklasnatter: `srcChainExpiryBlock`

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,323 @@
...
+        ...(failedSwap && { failure: getDepositIgnoredFailedState(failedSwap) }),
+      },
+      swap: {
+        ...rolledSwaps,
+        totalAmountSwapped: rolledSwaps?.totalAmountSwapped.toFixed(),
+        lastExecutedChunk:
+          rolledSwaps?.lastExecutedChunk && getSwapFields(rolledSwaps.lastExecutedChunk),
+        currentChunk: rolledSwaps && getSwapFields(rolledSwaps.currentChunk),
+        ...swapEgressFields,
+        type: sortedSwaps?.[0].type,
```
niklasnatter, line 288: think i would not include this for now - or is it useful?

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,323 @@
...
+        fees: aggregateFees,
+        srcChainRequiredBlockConfirmations,
+        estimatedDurationSeconds,
```
niklasnatter, line 291: this includes the witness time for the deposit - so i think we should move it to the root level

packages/swap/src/routes/v2/swap.ts
```diff
@@ -0,0 +1,323 @@
...
+          readField(swapRequest, failedSwap, 'depositAmount')?.toFixed() ?? pendingDeposit?.amount,
+        transactionRef: depositTransactionRef,
+        txConfirmations: pendingDeposit?.transactionConfirmations,
+        receivedAt: swapRequest?.depositReceivedAt?.valueOf(),
+        receivedBlockIndex: swapRequest?.depositReceivedBlockIndex ?? undefined,
+        ...(failedSwap && { failure: getDepositIgnoredFailedState(failedSwap) }),
+      },
+      swap: {
+        ...rolledSwaps,
+        totalAmountSwapped: rolledSwaps?.totalAmountSwapped.toFixed(),
```
niklasnatter, line 283: its not clear if this is the srcAsset or the destAsset. maybe we should have a `inputAmountSwapped` and a `outputAmountSwapped`?

niklasnatter: also fine for me to call it `totalInputAmountSwapped`/`totalOutputAmountSwapped` so the prefix matches with `totalChunksExecuted`

niklasnatter (commented):
i like it! sorry for all the comments, but it should be small stuff mostly

PR #905, opened 2024-09-04
feat(swap): handle failed ccms [WEB-1420]

acdibble:


packages/swap/src/event-handlers/index.ts
```diff
@@ -134,7 +135,7 @@ export const swapEventNames = Object.values(events).flatMap((pallets) => Object.
 
 export type EventHandlerArgs = {
   prisma: Prisma.TransactionClient;
-  event: Pick<Event, 'args' | 'name' | 'indexInBlock'>;
+  event: Pick<Event, 'args' | 'name' | 'indexInBlock' | 'callId'>;
   block: Pick<Block, 'height' | 'hash' | 'timestamp' | 'specId'>;
 };
```
GabrielBuragev, line 141: we probably need to attach the handler underneath

acdibble: every time

GabrielBuragev (approved):

PR #915, opened 2024-09-09
feat(swap): [WEB-1370] handle DCA quoting

jerryafr:


packages/swap/src/routes/quote.ts
```diff
@@ -38,6 +38,117 @@ const handleQuotingError = (res: express.Response, err: unknown, info: Additiona
   res.status(500).json({ message });
 };
 
+export const getDcaQuoteParams = async (asset: InternalAsset, amount: bigint) => {
+  const usdChunkSize = env.DCA_USD_CHUNK_SIZE;
+
+  const usdValue = await getUsdValue(amount, asset).catch(() => undefined);
+  if (usdValue) {
```
zoheb391, line 45:
```suggestion
  if (!usdValue) {
  logger.error(...)
  return null
  }
```

zoheb391:
i think you missed this part in the ticket: 

```
We should also disregard the quote if the total quote requested is greater than the total sum of liquidity within chainflip (active + inactive)
```

packages/swap/src/routes/quote.ts
```diff
@@ -38,6 +38,117 @@ const handleQuotingError = (res: express.Response, err: unknown, info: Additiona
   res.status(500).json({ message });
 };
 
+export const getDcaQuoteParams = async (asset: InternalAsset, amount: bigint) => {
+  const usdChunkSize = env.DCA_USD_CHUNK_SIZE;
```
zoheb391, line 42: lets create a map of asset to `usdChunksize` so we can change it easily on the go per asset in the future.

packages/swap/src/utils/getPoolQuote.ts
```diff
@@ -38,6 +39,7 @@ export default async function getPoolQuote({
   limitOrders?: LimitOrders;
   boostFeeBps?: number;
   pools: Pool[];
+  quoteType?: QuoteType;
```
niklasnatter, line 42: not sure if we need this here if we just pass it through

packages/swap/src/config/env.ts
```diff
@@ -43,6 +43,26 @@ export default z
     MAINTENANCE_MODE: optionalBoolean,
     LIQUIDITY_WARNING_THRESHOLD: optionalNumber(-5),
     COINGECKO_API_KEY: z.string().optional(),
+    DCA_USD_CHUNK_SIZE: optionalString('{}').transform((string) => {
+      try {
+        return z
```
niklasnatter, line 48: maybe an object would be easier to handle, but non strong opinion
```
{
  Btc: 2000,
  SolUsd: 500
}
```

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,334 @@
...
+  if (Number(usdValue) < usdChunkSize) {
+    return {
+      chunkSize: amount,
+      lastChunkAmount: amount,
+      numberOfChunks: 1,
+      addedDurationSeconds: 0,
+    };
+  }
+  const numberOfChunksPrecise = Number(usdValue) / usdChunkSize;
+  const chunkSize = BigNumber(amount.toString()).dividedBy(numberOfChunksPrecise);
```
niklasnatter, line 61: i think we set a `numberOfChunks` when creating a channel and then the protocol will just divide the amount by the number of chunks

niklasnatter: so i think we dont need the `numberOfChunksPrecise` or `lastChunkAmount` because all of the chunks will have the same size

packages/swap/src/config/env.ts
```diff
@@ -43,6 +43,26 @@ export default z
...
+      } catch (err) {
+        const error = err as Error;
+        // eslint-disable-next-line no-console
+        console.warn({
+          message: `Could not parse DCA_USD_CHUNK_SIZE variable. error: "${error?.message}"`,
+        });
+        return undefined;
+      }
+    }),
+    DCA_CHUNK_INTERVAL_SECONDS: optionalNumber(2),
```
niklasnatter, line 65: think the protocol uses a number of blocks instead of seconds

```suggestion
    DCA_CHUNK_INTERVAL_BLOCKS: optionalNumber(2),
```

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,334 @@
...
+
+  const lastChunkAmount = BigInt(
+    new BigNumber(amount.toString()).minus(chunkSize.multipliedBy(numberOfChunks - 1)).toFixed(0),
+  );
+
+  return {
+    chunkSize: BigInt(chunkSize.toFixed(0)),
+    lastChunkAmount,
+    numberOfChunks,
+    addedDurationSeconds: Math.ceil(env.DCA_CHUNK_INTERVAL_SECONDS * (numberOfChunks - 1)), // we deduct 1 chunk because the first one is already accounted for in the regular quote
```
niklasnatter, line 76: think the protocol uses a number of blocks instead of seconds, so we need to multiply by block time

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,334 @@
...
+        .multipliedBy(dcaQuoteParams.numberOfChunks - 1)
+        .plus(BigNumber(dcaBoostedQuote.egressAmount).multipliedBy(lastChunkRatio))
+        .toFixed(0),
+      estimatedDurationSeconds:
+        dcaBoostedQuote.estimatedDurationSeconds + dcaQuoteParams.addedDurationSeconds,
+    };
+  }
+};
+
+const fallbackChains = {
```
niklasnatter, line 144: think we can reuse this

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,334 @@
...
+  dcaQuote,
+  dcaBoostedQuote,
+  estimatedBoostFeeBps,
+}: {
+  dcaQuoteParams: NonNullable<Awaited<ReturnType<typeof getDcaQuoteParams>>>;
+  dcaQuote: QuoteQueryResponse;
+  dcaBoostedQuote?: QuoteQueryResponse | null | 0;
+  estimatedBoostFeeBps?: number;
+}) => {
+  const lastChunkRatio = new BigNumber(dcaQuoteParams.lastChunkAmount.toString()).dividedBy(
```
niklasnatter, line 91: calcs in here will be a bit easier if we assume that all chunks will be the same size

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    // eslint-disable-next-line no-param-reassign
+    dcaQuote.egressAmount = new BigNumber(dcaQuote.egressAmount)
+      .multipliedBy(dcaQuoteParams.numberOfChunks - 1)
+      .plus(BigNumber(dcaQuote.egressAmount).multipliedBy(lastChunkRatio))
+      .toFixed(0);
+    // eslint-disable-next-line no-param-reassign
+    dcaQuote.estimatedDurationSeconds += dcaQuoteParams.addedDurationSeconds;
+  }
```
niklasnatter, line 112: do we also need to scale the broker fee?

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,334 @@
...
+
+      return;
+    }
+
+    next();
+  });
+
+  router.get(
+    '/',
+    asyncHandler(async (req, res) => {
```
niklasnatter, line 169: i think a lot of the code in the handler is duplicated, right? maybe we could reuse the same handler with a query param to enable dca quotes?

niklasnatter: so if the query param is set we return an array, otherwise we keep the previous behaviour

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,334 @@
...
+              ...quoteArgs,
+              swapInputAmount: dcaQuoteParams.chunkSize,
+            }),
+          dcaQuoteParams &&
+            estimatedBoostFeeBps &&
+            getPoolQuote({
+              ...quoteArgs,
+              boostFeeBps: estimatedBoostFeeBps,
+              swapInputAmount: dcaQuoteParams.chunkSize,
+            }).catch(() => undefined),
```
niklasnatter, line 261: think it would be nicer to pass the quote to `adjustDcaQuote` here insie of `.then(adjustDcaQuote)`

niklasnatter: would need to adjust `adjustDcaQuote` to accept a single quote, but i think that would make sense in general?

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,334 @@
...
+    numberOfChunks,
+    addedDurationSeconds: Math.ceil(env.DCA_CHUNK_INTERVAL_SECONDS * (numberOfChunks - 1)), // we deduct 1 chunk because the first one is already accounted for in the regular quote
+  };
+};
+
+const adjustDcaQuote = ({
+  dcaQuoteParams,
+  dcaQuote,
+  dcaBoostedQuote,
+  estimatedBoostFeeBps,
```
niklasnatter, line 84: not sure if we need `estimatedBoostFeeBps` - cant we just scale the boost fee if it is there and ignore if it is not?

packages/swap/src/routes/v2/quote.ts
jerryafr: I thought about it, but I feel like it'd be strange if the return type would change so drastically (from object to array of objects) based on the params.
I think it's better to switch to a new v2 endpoint to always return an array of quotes

packages/swap/src/routes/v2/quote.ts
jerryafr: `dcaBoostedQuote` doesn't already contain `estimatedBoostFeeBps`. So when we're populating `dcaQuote.boostQuote`, we need to also use this value.

niklasnatter: okay i see - i think we should set the `estimatedBoostFeeBps` in `getPoolQuote` then 🙂 it would make the existing code a bit easier and that method already adds the fee to the fee array, so it already has all the data it needs

packages/swap/src/routes/v2/quote.ts
niklasnatter: alright fine for me too 🙂 but it would be nice to reuse the code if possible to not diverge - maybe we can extract a method that returns the array and then only stringify the first one in the `v1` handler?

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,325 @@
...
+    throw ServiceError.badRequest('insufficient liquidity for requested amount');
+  }
+
+  logger.error('error while collecting quotes:', err);
+
+  return message;
+};
+
+export const getDcaQuoteParams = async (asset: InternalAsset, amount: bigint) => {
+  const usdChunkSize = env.DCA_USD_CHUNK_SIZE?.[asset] ?? 3000;
```
zoheb391, line 43:
```suggestion
  const usdChunkSize = env.DCA_USD_CHUNK_SIZE?.[asset] ?? DEFAULT_CHUNK_SIZE_USD;
```

packages/shared/src/schemas.ts
```diff
@@ -147,4 +147,8 @@ export type BoostedQuoteDetails = QuoteDetails & { estimatedBoostFeeBps: number
 
 export type QuoteQueryResponse = QuoteDetails & {
   boostQuote?: BoostedQuoteDetails;
+  dcaParams?: {
+    numberOfChunks: number;
+    chunkInterval: number;
```
zoheb391, line 152:
```suggestion
    chunkIntervalBlocks: number;
```

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,325 @@
...
+  if (Number(usdValue) < usdChunkSize) {
+    return {
+      chunkSize: amount,
+      numberOfChunks: 1,
+      addedDurationSeconds: 0,
+    };
+  }
+  const numberOfChunksPrecise = Number(usdValue) / usdChunkSize;
+  const numberOfChunks =
+    numberOfChunksPrecise - Math.floor(numberOfChunksPrecise) < 0.05
```
niklasnatter, line 61: not sure if this is needed, but not a strong opinion

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,325 @@
...
+
+        if (dcaQuoteParams && dcaQuote) {
+          // Check liquidity for DCA
+          if (srcAsset === 'Usdc' || destAsset === 'Usdc') {
+            const totalLiquidity = await getTotalLiquidity(srcAsset, destAsset);
+            if (
+              totalLiquidity <
+              BigInt(dcaQuote.egressAmount) * BigInt(dcaQuoteParams.numberOfChunks)
+            ) {
+              throw ServiceError.badRequest(`Insufficient liquidity for the requested amount 1`);
```
zoheb391, line 263:
```suggestion
              throw ServiceError.badRequest(`Insufficient liquidity for the requested amount `);
```

zoheb391: same below

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,325 @@
...
+              ...quoteArgs,
+              swapInputAmount: dcaQuoteParams.chunkSize,
+            }),
+          dcaQuoteParams &&
+            estimatedBoostFeeBps &&
+            getPoolQuote({
+              ...quoteArgs,
+              boostFeeBps: estimatedBoostFeeBps,
+              swapInputAmount: dcaQuoteParams.chunkSize,
+            }).catch(() => undefined),
```
zoheb391, line 252: prob i don't understand how boost quoting works but why do we need to make this extra call here? can't we just take  pass `estimatedBoostFeeBps` to the adjust function.

packages/shared/src/schemas.ts
zoheb391: actually nvm, since the FE passes this response to the openDC method lets keep it  so no transformations are necessary.

niklasnatter (approved):

acdibble (approved):

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,370 @@
...
+      egressAmount: BigNumber(dcaBoostedQuote.egressAmount)
+        .multipliedBy(dcaQuoteParams.numberOfChunks)
+        .toFixed(0),
+      estimatedDurationSeconds:
+        dcaBoostedQuote.estimatedDurationSeconds + dcaQuoteParams.addedDurationSeconds,
+    };
+  }
+};
+
+export const validateQuoteQuery = async (query: Query) => {
```
zoheb391, line 142: i would move these shared functions into its own file

zoheb391 (approved):

packages/swap/src/routes/v2/quote.ts
jerryafr: as discussed offline, will do this in a later PR

packages/swap/src/routes/v2/quote.ts
```diff
@@ -0,0 +1,370 @@
...
+  const numberOfChunks = Math.ceil(Number(usdValue) / usdChunkSize);
+
+  return {
+    chunkSize: BigInt(new BigNumber(amount.toString()).dividedBy(numberOfChunks).toFixed(0)),
+    numberOfChunks,
+    addedDurationSeconds: Math.ceil(env.DCA_CHUNK_INTERVAL_BLOCKS * 6 * (numberOfChunks - 1)), // we deduct 1 chunk because the first one is already accounted for in the regular quote
+  };
+};
+
+const adjustDcaQuote = ({
```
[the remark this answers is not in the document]
jerryafr: network fee is handled here, I don't think we would need to do anything else here.
will keep that in my mind though during testing to ensure nothing is missed

PR #919, opened 2024-09-11
chore: (swap, sdk):  add types and address pr comments

zoheb391:
types are still WIP

changing them as i'm going


packages/swap/src/routes/v2/swap.ts
```diff
@@ -235,22 +237,28 @@ router.get(
       internalSrcAsset && getRequiredBlockConfirmations(internalSrcAsset),
     ]);
 
+    const showCcm = Boolean(swapDepositChannel?.ccmGasBudget || swapDepositChannel?.ccmMessage);
+    const showRefund = swapDepositChannel?.fokRefundAddress;
+    const showEgress = Object.keys(swapEgressFields).length !== 0;
+
     const response = {
```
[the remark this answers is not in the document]
zoheb391: not sure what you mean

packages/sdk/src/swap/sdk.ts
```diff
@@ -238,7 +238,7 @@ export class SwapSDK {
   getStatus(
     swapStatusRequest: SwapStatusRequest,
     options: ApiService.RequestOptions = {},
-  ): Promise<SwapStatusResponse> {
+  ): Promise<SwapV2.SwapStatusResponse> {
```
niklasnatter, line 241: i think we need to do a separate `getStatusV2` method and keep the behaviour of this one for backwards compatibility

packages/sdk/src/swap/v2/types.ts
```diff
@@ -0,0 +1,183 @@
...
+  type CcmParams,
+  type DcaParams as Dca,
+} from '@/shared/schemas';
+import { SwapType } from '@/swap/client';
+import { FailureMode } from '@/swap/utils/swap';
+import { ChainsAndAssets } from '../types';
+
+// eslint-disable-next-line @typescript-eslint/no-namespace
+export namespace SwapV2 {
+  type DcaParameters = Omit<Dca, 'chunkInterval'> & {
```
niklasnatter, line 14: think i would be explicit here and add all of the properties explicitly instead of combining types

packages/sdk/src/swap/v2/types.ts
```diff
@@ -0,0 +1,183 @@
...
+    chunkIntervalBlocks: string;
+  };
+
+  type Ccm = CcmParams & {
```
niklasnatter, line 18: same here

packages/sdk/src/swap/v2/types.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  type WithEgressFields<T, F = undefined> = T & {
+    egress: (F extends undefined ? EgressFields : EgressFields & { failure: F }) | undefined;
+  };
+  type WtihEgressSentFields<T, F = undefined> = T & {
+    egress:
+      | (F extends undefined ? EgressSentFields : EgressSentFields & { failure: F })
+      | undefined;
+  };
+  type WtihEgressCompletedFields<T> = T & {
```
niklasnatter, line 120: `With`

packages/sdk/src/swap/v2/types.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  type SendingSwapFields = WithEgressFields<SwapFields, Failure> | undefined;
+  type SentSwapFields = WtihEgressSentFields<SwapFields, Failure> | undefined;
+  type CompletedSwapFields = WtihEgressCompletedFields<SwapFields> | undefined;
+
+  type SendingRefundFields = WithFailure<RefundFields & EgressFields> | undefined;
+  type SentRefundFields = WithFailure<RefundFields & EgressSentFields> | undefined;
+  type CompletedRefundFields = EgressCompletedFields | undefined;
+
+  interface DepositChannel extends SwapStatusResponseCommonFields {
```
niklasnatter, line 132: just curious - is there a reason for using `interface` here but `type` above?

packages/swap/src/routes/v2/swap.ts
```diff
@@ -272,23 +280,21 @@ router.get(
...
+        totalOutputAmountSwapped: rolledSwaps?.totalOutputAmountSwapped.toFixed(),
         lastExecutedChunk:
           rolledSwaps?.lastExecutedChunk && getSwapFields(rolledSwaps.lastExecutedChunk),
         currentChunk: rolledSwaps && getSwapFields(rolledSwaps.currentChunk),
-        ...swapEgressFields,
-        type: sortedSwaps?.[0].type,
         fees: aggregateFees,
-        srcChainRequiredBlockConfirmations,
-        estimatedDurationSeconds,
+        ...(showEgress && { egress: { ...swapEgressFields } }),
```
niklasnatter, line 297: i think i would move this out of the `swap` so its on the same level as `refund`

niklasnatter: then we would have `deposit`, `swap`, `egress`, `refund` on that level which feels right to me (?)

packages/sdk/src/swap/v2/types.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  type EgressFields = {
+    amount: string;
+    scheduledAt: number;
+    scheduledBlockIndex: string;
+  };
+  type EgressSentFields = EgressFields & {
+    sentTxRef: string;
+  };
+  type EgressCompletedFields = EgressFields & {
```
niklasnatter, line 104: while it is correct to have different types for a scheduled and completed egress, i am not sure if its worth it. i would be fine with just having one type:
```
  type EgressFields = {
    amount: string;
    scheduledAt: number;
    scheduledBlockIndex: string;
    sentTxRef: string | undefined;
    sentAt: number | undefined;
    sentAtBlockIndex: string | undefined;
  };
```

PR #921, opened 2024-09-12
chore(sdk): simplify types and add status v2 method

zoheb391:


packages/sdk/src/swap/v2/types.ts
```diff
@@ -72,89 +65,52 @@ export namespace SwapV2 {
...
-    depositChannel: DepositChannelFields;
-  }
-  interface Receiving extends DepositChannel {
-    deposit: WithFailure<DepositFields>;
-  }
-  interface Swapping extends Receiving {
+  type Swapping = Receiving & {
     swap: SwapFields;
-  }
-  interface Sending extends Receiving {
```
acdibble, line 141: maybe extending `interface`s provides a better DX? usually type intersections provide some janky, hard to read types

packages/sdk/src/swap/sdk.ts
```diff
@@ -238,10 +239,17 @@ export class SwapSDK {
   getStatus(
     swapStatusRequest: SwapStatusRequest,
     options: ApiService.RequestOptions = {},
-  ): Promise<SwapV2.SwapStatusResponse> {
+  ): Promise<SwapStatusResponse> {
     return ApiService.getStatus(this.options.backendUrl, swapStatusRequest, options);
   }
 
+  getStatusV2(
```
niklasnatter, line 246: could do something like this until we know the shape will stah the same. but no strong opinion
```suggestion
  /** @experimental */
  getStatusV2(
```

niklasnatter (approved):

PR #931, opened 2024-09-16
refactor(swap): add fields to response

acdibble:


jerryafr (approved):

packages/sdk/src/swap/v2/types.ts
```diff
@@ -83,13 +83,16 @@ interface EgressFields {
...
+  failedAt: number | undefined;
+  failedAtBlockIndex: string | undefined;
 }
 interface SwapStatusResponseCommonFields extends ChainsAndAssets {
   swapId: string;
   destAddress: string;
   ccm: Ccm | undefined;
   boost: Boost | undefined;
   estimatedDurationSeconds: number | null | undefined;
+  srcChainRequiredBlockConfirmations: number | null;
```
acdibble, line 95: moved this here because it's true for vault swaps too (even if we don't currently create a swap record until it's confirmed)

packages/sdk/src/swap/v2/types.ts
```diff
@@ -63,12 +62,13 @@ interface SwapChunk {
   latestSwapRescheduledAt: number | undefined;
   latestSwapRescheduledBlockIndex: number | undefined;
   fees: SwapFee[];
+  isDca: boolean;
```
niklasnatter, line 65: already have this on the swap - do we need it on the chunk too?

packages/sdk/src/swap/v2/types.ts
```diff
@@ -83,13 +83,16 @@ interface EgressFields {
   sentAtBlockIndex: string | undefined;
```
niklasnatter, line 83: think this is `confirmedAt`

packages/swap/src/routes/v2/utils.ts
```diff
@@ -127,19 +127,26 @@ export const getEgressStatusFields = async (
...
       amount: egress.amount?.toFixed(),
       scheduledAt: egress.scheduledAt?.valueOf(),
       scheduledBlockIndex: egress.scheduledBlockIndex ?? undefined,
     }),
     ...(broadcast && {
-      sentAt: broadcast?.succeededAt?.valueOf(),
-      sentAtBlockIndex: broadcast?.succeededBlockIndex ?? undefined,
+      confirmedAt: broadcast?.succeededAt?.valueOf(),
+      confirmedBlockIndex: broadcast?.succeededBlockIndex ?? undefined,
       sentTxRef: broadcast?.transactionRef ?? egressTrackerTxRef,
```
niklasnatter, line 140: could just call this `txRef` now that we renamed `sent`

acdibble: done

packages/sdk/src/swap/v2/types.ts
acdibble: yeah because if you access swap the type is `SwapFields | SwapChunk | (SwapFields & SwapChunk) | undefined` so it's a bit annoying not being able to access the `isDca` field in all cases

niklasnatter (approved):

packages/swap/src/routes/__tests__/v2/swap.test.ts
```diff
@@ -1805,86 +1413,7 @@ describe('server', () => {
...
-            mode: 'REFUND_OUTPUT_TOO_SMALL',
-            reason: {
-              name: 'ethereumIngressEgress.BelowEgressDustLimit',
-              message: 'The amount is below the minimum egress amount.',
-            },
-          },
-          ignoredAmount: '999844999999160000',
-        },
-      });
+      expect(rest).toMatchSnapshot();
```
zoheb391, line 1416: not a big fan of these empty methods. adding the expected props gives us two layers of protection. Otherwise I feel like you can easily end up updating snapshots without realising it.

packages/swap/src/routes/v2/swap.ts
```diff
@@ -310,8 +315,8 @@ router.get(
             }),
         fees: aggregateFees,
       },
-      ...(showSwapEgress && { swapEgress: { ...swapEgressFields } }),
-      ...(showRefundEgress && { refundEgress: { ...refundEgressFields } }),
+      ...(swapEgressFields && { swapEgress: { ...swapEgressFields } }),
+      ...(refundEgressFields && { refundEgress: { ...refundEgressFields } }),
```
zoheb391, line 319: i think you could have empty objects now.

zoheb391 (approved):

packages/swap/src/routes/__tests__/v2/swap.test.ts
acdibble: Yeah it's just a pain to have to double update stuff. Ideally we should all be reviewing snapshot changes in PRs. Otherwise we should just assert that the object strictly equals

packages/swap/src/routes/v2/swap.ts
acdibble: We shouldn't, I changed the function to return null if none of the data is present

PR #933, opened 2024-09-17
refactor(swap): add waiting state

acdibble:


jerryafr (approved):

packages/sdk/src/swap/v2/types.ts
```diff
@@ -140,6 +142,9 @@ interface Sending extends Receiving {
 }
 
 type SwapState =
+  | ({
+      state: 'PENDING';
```
niklasnatter, line 146: not sure if pending is the best term here - what do you think about `READY`/`OPEN`/`CREATED`?

niklasnatter: or `REQUESTED`? bc the method to open the channel is called `requestDepositAddress`

niklasnatter (approved):

PR #935, opened 2024-09-17
feat(sdk): [WEB-1370] add getQuoteV2 to sdk

jerryafr:


packages/sdk/src/swap/sdk.ts
```diff
@@ -137,6 +138,7 @@ export class SwapSDK {
       .filter((asset) => !chain || asset.chain === chain);
   }
 
+  /** @deprecated DEPRECATED(1.6) use getQuoteV2() */
```
niklasnatter, line 141: might need to add this to the changelog

packages/sdk/src/swap/sdk.ts
```diff
@@ -157,6 +159,26 @@ export class SwapSDK {
     );
   }
 
+  getQuoteV2(
```
niklasnatter, line 162: would be nice to copy the tests for this

zoheb391 (approved):
you're missing a test in sdk.test.ts

PR #937, opened 2024-09-18
chore(swap): Adjust ccm type in status v2

niklasnatter:


packages/sdk/src/swap/v2/types.ts
```diff
@@ -8,11 +8,9 @@ interface DcaParameters {
   chunkIntervalBlocks: string;
 }
 
-interface Ccm {
+interface CcmParameters {
   message: string;
   gasBudget: string;
-  cfParameters: string | undefined;
-  ccmDepositReceivedBlockIndex: string | undefined;
```
niklasnatter, line 15: i think is no separate ccm deposit event after the events refactor

packages/sdk/src/swap/v2/types.ts
```diff
@@ -8,11 +8,9 @@ interface DcaParameters {
   chunkIntervalBlocks: string;
 }
 
-interface Ccm {
+interface CcmParameters {
   message: string;
   gasBudget: string;
-  cfParameters: string | undefined;
```
niklasnatter, line 14: afaik these are not used yet, we can still add them once there is a reason

packages/sdk/src/swap/v2/types.ts
```diff
@@ -101,7 +99,7 @@ interface EgressFields {
 interface SwapStatusResponseCommonFields extends ChainsAndAssets {
   swapId: string;
   destAddress: string;
-  ccm: Ccm | undefined;
+  ccmParams: CcmParameters | undefined;
```
niklasnatter, line 102: makes the property name consistent with `requestDepositAddress`

packages/sdk/src/swap/v2/types.ts
```diff
@@ -114,10 +112,7 @@ interface ReceivingVaultSwap extends VaultSwapCommonFields {
 }
 
 interface SwappingVaultSwap extends ReceivingVaultSwap {
-  swap:
-    | Exclude<SwapChunk, 'latestSwapRescheduledAt' | 'latestSwapRescheduledBlockIndex'>
-    | undefined;
-  swapEgress: EgressFields | undefined;
+  swap: SwapChunk | undefined;
```
niklasnatter, line 115: contract swaps can be rescheduled if the global price impact protection hits

niklasnatter: not sure if we gain anything from separate contract swaps from channel swaps in the types now, but can revisit that later

packages/sdk/src/swap/v2/types.ts
```diff
@@ -133,11 +128,11 @@ interface Receiving extends DepositChannel {
 }
 
 interface Swapping extends Receiving {
-  swap: SwapFields;
+  swap: SwapFields | SwapChunk;
```
niklasnatter, line 131: channel swaps will return `SwapChunk` if the swap is not a dca swap atm

packages/sdk/src/swap/v2/types.ts
```diff
@@ -81,7 +79,6 @@ interface SwapFields {
   isDca: true;
   lastExecutedChunk: SwapChunk | undefined;
   currentChunk: SwapChunk;
-  type: SwapType;
```
niklasnatter, line 84: never set, but also think its fine to not expose this

packages/sdk/src/swap/v2/types.ts
```diff
@@ -92,7 +89,6 @@ interface EgressFields {
   txRef: string | undefined;
   confirmedAt: number | undefined;
   confirmedBlockIndex: string | undefined;
-  ignoredAmount: string | undefined;
```
niklasnatter, line 95: use `amount` and `failure` if egress was ignored (consistent with deposit)

acdibble (approved):

jerryafr (approved):

PR #939, opened 2024-09-18
chore(swap): Remove redundant vault swap type

niklasnatter:


packages/sdk/src/swap/v2/types.ts
```diff
@@ -124,10 +111,10 @@ interface Sending extends Receiving {
   refundEgress: EgressFields | undefined;
 }
 
-type SwapState =
+export type SwapStatusResponseV2 =
```
niklasnatter, line 114: as far as i understand, this does not change the resulting type for `SwapStatusResponseV2` at all

packages/sdk/src/swap/v2/types.ts
```diff
@@ -86,31 +86,18 @@ interface EgressFields {
...
   swapId: string;
   destAddress: string;
+  depositChannel: DepositChannelFields | undefined;
   ccmParams: CcmParams | undefined;
   boost: Boost | undefined;
   estimatedDurationSeconds: number | null | undefined;
   srcChainRequiredBlockConfirmations: number | null;
 }
 
-interface VaultSwapCommonFields extends Exclude<SwapStatusResponseCommonFields, 'boost'> {}
```
acdibble, line 95: technically vault swaps can't be boost or dca but yeah i'm not sure if we gain much by making it explicit in types

acdibble (approved):

packages/sdk/src/swap/v2/types.ts
niklasnatter: true! but i think we should communicate that via the docs rather than the types if it allows us the make the types easier to understand

packages/sdk/src/swap/v2/types.ts
```diff
@@ -86,31 +86,18 @@ interface EgressFields {
 interface SwapStatusResponseCommonFields extends ChainsAndAssets {
   swapId: string;
   destAddress: string;
+  depositChannel: DepositChannelFields | undefined;
```
zoheb391, line 89: this is already extended in waiting. we can remove it here. Or remove it from waiting

niklasnatter: i marked it as mandatory in `WAITING` because that status can only happen for deposit channel swaps. thats why we need it nullable here 🙂

PR #940, opened 2024-09-18
chore(swap): Expose witnessedAt for deposits and egresses

niklasnatter:


packages/sdk/src/swap/v2/types.ts
```diff
@@ -76,8 +76,8 @@ interface EgressFields {
   scheduledAt: number;
   scheduledBlockIndex: string;
   txRef: string | undefined;
-  confirmedAt: number | undefined;
```
niklasnatter, line 79: `confirmedAt` is a bit ambiguous because its also called `confirmed` when a transaction is included in a block

niklasnatter: i think `witnessedAt` is a nice alternative bc you find something helpful when you google for `chainflip witness` and we can also use it for deposits

jerryafr (approved):

packages/sdk/src/swap/v2/types.ts
acdibble: but we also have a concept of # of confirmations required

acdibble (approved):

PR #941, opened 2024-09-18
refactor(swap): update swap shape

acdibble:


packages/sdk/src/swap/v2/types.ts
```diff
@@ -47,27 +47,39 @@ interface DepositFields {
...
+      executedAt: number;
+      executedBlockIndex: string;
+      retryCount: number;
+    } | null;
+    currentChunk: {
+      inputAmount: string;
+      scheduledAt: number;
+      scheduledBlockIndex: string;
+      retryCount: number;
+    } | null;
```
niklasnatter, line 79: not sure if this can be null

packages/swap/src/routes/v2/swap.ts
```diff
@@ -149,32 +148,36 @@ router.get(
...
-      if (a.swapExecutedAt && b.swapExecutedAt) {
-        return a.swapExecutedAt.valueOf() - b.swapExecutedAt.valueOf() ? 1 : -1;
-      }
-      return a.swapExecutedAt && !b.swapExecutedAt ? 1 : 0;
-    });
+    let originalInputAmount;
+
+    if (swapRequest) {
+      originalInputAmount = swapRequest.fees.reduce(
+        (acc, fee) => (fee.asset === swapRequest.srcAsset ? acc.minus(fee.amount) : acc),
```
niklasnatter, line 155: might need to filter to only subtract ingress fee

packages/swap/src/routes/v2/swap.ts
```diff
@@ -183,6 +186,7 @@ router.get(
     );
 
     const aggregateFees = rolledSwaps?.fees
+      .splice(0)
```
niklasnatter, line 189: 🤔

niklasnatter (approved):
looks nice 🙌

zoheb391 (approved):
think we're ready for beta?

packages/swap/src/routes/v2/swap.ts
acdibble: for older swaps the broker fee too

packages/sdk/src/swap/v2/types.ts
acdibble: it can be null if the entire DCA swap is completed

PR #946, opened 2024-09-19
refactor(swap): store original swap input and remove old handlers

acdibble:


packages/swap/prisma/migrations/20240919124938_add_swap_input_amount/migration.sql
acdibble, line 1: important

packages/swap/src/event-handlers/swapRequested.ts
```diff
@@ -121,6 +121,7 @@ export default async function swapRequested({
       srcAddress: ccmMetadata?.sourceAddress?.address,
       destAddress,
       swapRequestedAt: new Date(block.timestamp),
+      swapInputAmount: inputAmount.toString(),
```
acdibble, line 124: important

packages/swap/src/event-handlers/ccmDepositReceived.ts
```diff
@@ -1,38 +0,0 @@
...
-export default async function ccmDepositReceived({ prisma, event, block }: EventHandlerArgs) {
-  const { principalSwapId, depositMetadata } = ccmDepositReceivedArgs.parse(event.args);
-
-  if (principalSwapId) {
-    await prisma.swapRequest.update({
-      where: {
-        nativeId: principalSwapId,
-      },
-      data: {
-        ccmDepositReceivedBlockIndex: `${block.height}-${event.indexInBlock}`,
```
niklasnatter, line 32: can we remove this field? thinks its not used anymore

niklasnatter (approved):

jerryafr (approved):

packages/swap/src/event-handlers/ccmDepositReceived.ts
acdibble: It's used in the status endpoint, we'd have to deprecated it first but we are already deprecating that endpoint anyway

PR #957, opened 2024-09-23
fix: use swap request timestamp for vault swaps

GabrielBuragev:


packages/swap/src/routes/v2/utils.ts
```diff
@@ -83,7 +83,10 @@ export const getDepositInfo = (
       amount,
       txRef: depositTransactionRef,
       txConfirmations: pendingDeposit?.transactionConfirmations,
-      witnessedAt: swapRequest?.depositReceivedAt?.valueOf(),
+      witnessedAt:
+        swapRequest?.originType === 'VAULT'
+          ? swapRequest.swapRequestedAt.valueOf()
+          : swapRequest?.depositReceivedAt?.valueOf(),
```
GabrielBuragev, line 89: We don't have a `depositReceivedAt` entry for smart contract swaps therefore defaulting to the `swapRequestedAt`

niklasnatter (approved):
test would be great to ensure we dont change

jerryafr (approved):

GabrielBuragev:
> test would be great to ensure we dont change

i think the updated snapshots should cover that. Let me know if you think i should test the util function itself. I think the swap response test is sufficient

PR #959, opened 2024-09-23
refactor(swap): add intermediate amount

acdibble:


packages/sdk/src/swap/v2/types.ts
```diff
@@ -51,10 +51,12 @@ interface SwapFields {
   originalInputAmount: string;
   remainingInputAmount: string;
   swappedInputAmount: string;
+  swappedIntermediateAmount: string;
   swappedOutputAmount: string;
   regular?: {
     inputAmount: string;
-    outputAmount: string;
+    intermediateAmount?: string;
```
niklasnatter, line 58: do we need to add to `dca` too?

niklasnatter (approved):

packages/sdk/src/swap/v2/types.ts
acdibble: yeah

PR #968, opened 2024-09-26
feat(swap): [WEB-1484] SDK quote to return broker fee always in USDC

jerryafr:


packages/swap/src/utils/statechain.ts
```diff
@@ -12,11 +13,67 @@ export type SwapRateArgs = {
...
+  }
+  let usdcAmount = intermediateAmount ?? inputAmount;
+  if (destAsset === 'Usdc') {
+    usdcAmount = outputAmount;
+  }
+
+  const brokerFee = BigInt(
+    brokerCommissionBps && getPipAmountFromAmount(usdcAmount, brokerCommissionBps),
+  );
+
```
acdibble, line 47: i would remove the broker fee up here because any following calculation will have to do it otherwise

packages/swap/src/utils/statechain.ts
```diff
@@ -12,11 +13,67 @@ export type SwapRateArgs = {
...
+  const outputAmountExchangeRate = (outputAmount + egressFee) / (intermediateAmount ?? inputAmount);
```
acdibble, line 48: like here the rate includes the broker fee still

acdibble (approved):

packages/swap/src/utils/statechain.ts
```diff
@@ -12,11 +13,67 @@ export type SwapRateArgs = {
...
+
+  if (intermediateAmount) {
+    return {
+      intermediateAmount: intermediateAmount - brokerFee,
+      outputAmount: outputAmount - brokerFee * outputAmountExchangeRate,
```
niklasnatter, line 53: not 100% sure if i understand the `outputAmountExchangeRate` - think we could also do something like this
```suggestion
      outputAmount: outputAmount - getPipAmountFromAmount(outputAmount, brokerCommissionBps),
```

packages/swap/src/utils/statechain.ts
```diff
@@ -12,11 +13,67 @@ export type SwapRateArgs = {
...
+      outputAmount: outputAmount - brokerFee,
+      brokerFee,
+    };
+  }
+  // Source asset is USDC
+  return {
+    outputAmount: outputAmount - brokerFee * outputAmountExchangeRate,
+    brokerFee,
+  };
+};
```
GabrielBuragev, line 68: Would be nice to have a test for this function so we can verify the numbers

packages/swap/src/utils/statechain.ts
jerryafr: thanks for the suggestion, simplified the logic based on that :)

packages/swap/src/utils/statechain.ts
jerryafr: updated the logic and made it simpler, please check again 🙏

niklasnatter (approved):
🙌

PR #969, opened 2024-09-26
fix(shared): pass cf params from new object

acdibble:


packages/shared/src/vault/executeSwap.ts
```diff
@@ -174,7 +174,7 @@ const callToken = async (
     params.ccmParams.gasBudget,
     erc20Address,
     params.amount,
-    params.ccmMetadata?.cfParameters ?? '0x',
+    params.ccmParams?.cfParameters ?? '0x',
```
acdibble, line 177: this is the fix

niklasnatter (approved):

PR #972, opened 2024-09-26
feat(swap): add new quoting auth method [WEB-1494]

acdibble:


packages/swap/src/quoting/Quoter.ts
```diff
@@ -97,11 +120,45 @@ export default class Quoter {
...
+    const quotedLegsMap = new Map<string, LegFormatter>();
+
+    for (const socket of await this.io.fetchSockets()) {
+      const [first, second] = request.legs;
+      const { quotedAssets, marketMaker } = socket.data;
+      const quotesFirstLeg = quotedAssets[first.getBaseAsset()];
+      const quotesSecondLeg = !second || quotedAssets[second.getBaseAsset()];
+      let message: MarketMakerQuoteRequest<LegJson>;
+
+      if (quotesFirstLeg && quotesSecondLeg) {
```
jerryafr, line 140: wouldn't `quotesSecondLeg` always either have value or be true?

GabrielBuragev (approved):

jerryafr (approved):

packages/swap/src/quoting/Quoter.ts
```diff
@@ -97,11 +120,45 @@ export default class Quoter {
...
-    this.io.emit('quote_request', request);
+    let expectedResponses = 0;
+
+    const quotedLegsMap = new Map<string, LegFormatter>();
+
+    for (const socket of await this.io.fetchSockets()) {
+      const [first, second] = request.legs;
+      const { quotedAssets, marketMaker } = socket.data;
+      const quotesFirstLeg = quotedAssets[first.getBaseAsset()];
+      const quotesSecondLeg = !second || quotedAssets[second.getBaseAsset()];
```
niklasnatter, line 137: i dont understand the `!second` here but im sure there is a reason

niklasnatter (approved):

packages/swap/src/quoting/Quoter.ts
acdibble: If there is no second leg

niklasnatter: but why do we need to set `quotesSecondLeg` to true then?

acdibble: So the first `if` statement below is used. Can probably simplify/make this more explicit

jerryafr (approved):

GabrielBuragev (approved):

niklasnatter (approved):
nice

PR #974, opened 2024-09-27
fix(swap): refine dca quote calculation for fees

jerryafr:


packages/swap/src/routes/v2/quote.ts
```diff
@@ -225,25 +227,61 @@ export const generateQuotes = async ({
...
+    )
+    .toString();
+
+  const [dcaQuoteResult, dcaBoostedQuoteResult] = await Promise.allSettled([
+    dcaQuoteParams &&
+      getPoolQuote({
+        ...quoteArgs,
+        swapInputAmount:
+          dcaQuoteParams.chunkSize +
+          BigInt(
```
niklasnatter, line 263: hmm this is hard to understand for me.. i think we could achieve the same in the `adjustDcaQuote` without quoting an adjusted amount, right?

niklasnatter: as far as i understand, the problem is that at the moment we do
```
dcaQuote.egressAmount = new BigNumber(dcaQuote.egressAmount)
      .multipliedBy(dcaQuoteParams.numberOfChunks)
```
which basically includes the fees `numberOfChunks` times, right?

niklasnatter: but we could calculate a `chunkOutputAmount = chunkSize * exchangeRate` and then doe something like this?

```
dcaQuote.egressAmount = new BigNumber(dcaQuote.egressAmount)
      .add(chunkOutputAmount * (dcaQuoteParams.numberOfChunks - 1))
```

zoheb391 (approved):
some comments explaining the shenanigans would be helpful.

packages/swap/src/routes/v2/quote.ts
```diff
@@ -225,25 +230,50 @@ export const generateQuotes = async ({
...
+  const ingressFee =
+    quoteResult.status === 'fulfilled'
+      ? quoteResult.value.includedFees.find((fee) => fee.type === 'INGRESS')
+      : undefined;
+  const egressFee =
+    quoteResult.status === 'fulfilled'
+      ? quoteResult.value.includedFees.find((fee) => fee.type === 'EGRESS')
+      : undefined;
+
+  const ingressFeeSurcharge = dcaQuoteParams
```
niklasnatter, line 246:
```suggestion
  // the swap_rate rpc will deduct the full ingress fee before simulating the swap
  // as we quote a single chunk, we add a surcharge so that the effective deducted amount is 1/numberOfChunks
  const ingressFeeSurcharge = dcaQuoteParams
```

packages/swap/src/routes/v2/quote.ts
```diff
@@ -225,25 +230,50 @@ export const generateQuotes = async ({
...
+    ? BigInt(
+        new BigNumber(ingressFee?.amount ?? 0)
+          .multipliedBy((dcaQuoteParams.numberOfChunks - 1) / dcaQuoteParams.numberOfChunks)
+          .toFixed(0),
+      )
+    : 0n;
+  const egressFeeSurcharge = dcaQuoteParams
```
niklasnatter, line 253: think we dont need to calculate this here but can do it inside the `adjustDcaQuote` (all data should be on the `dcaQuote` already)

packages/swap/src/routes/v2/quote.ts
```diff
@@ -96,6 +98,7 @@ const adjustDcaQuote = ({
     // eslint-disable-next-line no-param-reassign
     dcaQuote.egressAmount = new BigNumber(dcaQuote.egressAmount)
```
niklasnatter, line 99: if we calculate the `egressFeeSurcharge` here, we can add a comment like
```
// when multiplying the egressAmount with numberOfChunks, we will deduct the egressFee multiple times
```

niklasnatter (approved):
im struggling to find a good explanation, so feel free to use something else if you have a better idea!

packages/shared/src/schemas.ts
```diff
@@ -137,14 +137,14 @@ export type QuoteDetails = {
   lowLiquidityWarning: boolean | undefined;
   estimatedDurationSeconds: number;
   estimatedPrice: string;
-};
-export type BoostedQuoteDetails = QuoteDetails & { estimatedBoostFeeBps: number };
-
-export type QuoteQueryResponse = QuoteDetails & {
-  boostQuote?: BoostedQuoteDetails;
   dcaParams?: {
```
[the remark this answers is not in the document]
jerryafr: actually I was thinking maybe we should remove type from boost as well haha
let's discuss this later :)

GabrielBuragev (approved):
Some comments/simplification would help with the readiness, took me some time to do the mental math. Great catch though!

PR #983, opened 2024-09-30
refactor: require FoK params for DCA channels

acdibble:


GabrielBuragev (approved):

packages/shared/src/schemas.ts
```diff
@@ -106,6 +120,8 @@ export const openSwapDepositChannelSchema = z
     fillOrKillParams: fillOrKillParams.optional(),
     dcaParams: dcaParams.optional(),
   })
+  .superRefine(ensureDcaWithFok)
+  // openSwapDepositChannelSchema
```
zoheb391, line 124: ?

acdibble, line 124:
```suggestion
```

packages/shared/src/__tests__/schema.test.ts
```diff
@@ -78,17 +78,56 @@ describe('postSwapSchema', () => {
...
+        fillOrKillParams: {
+          retryDurationBlocks: 10,
+          refundAddress: '0x1234',
+          minPriceX128: '1',
+        },
+      }),
+    ).not.toThrow();
+  });
+
+  it('only allows FoK params without DCA params', () => {
```
zoheb391, line 121:
```suggestion
  it('allows FoK params without DCA params', () => {
```

zoheb391 (approved):

niklasnatter:
> Is this a protocol level requirement or only enforced by the SDK?

It is enforced by the SDK for now, but it might be added on the protocol level in the future.

CumpsD:
> It is enforced by the SDK for now, but it might be added on the protocol level in the future.

Thanks!

PR #985, opened 2024-10-01
fix(sdk): Use swapInput and swapOutput to calculate estimatedRate

niklasnatter:


packages/swap/src/utils/getPoolQuote.ts
```diff
@@ -25,7 +25,7 @@ const getPrice = (
 export default async function getPoolQuote({
   srcAsset,
   destAsset,
-  swapInputAmount: originalSwapInputAmount,
```
niklasnatter, line 28: i think we initially deducted the ingress fee outside this method. but then with `swap_rate_v2` the ingress fee calculation moved into the rpc and therefore the value we pass here is the deposit amount

packages/swap/src/utils/getPoolQuote.ts
```diff
@@ -114,7 +118,7 @@ export default async function getPoolQuote({
       destAsset,
       boosted: Boolean(boostFeeBps),
     }),
-    estimatedPrice: getPrice(swapInputAmount, srcAsset, egressAmount, destAsset),
+    estimatedPrice: getPrice(swapInputAmount, srcAsset, swapOutputAmount, destAsset),
```
niklasnatter, line 121: the `estimatedPrice` should be the price at which the assets are swapped at the amm level to allow for calculating a fok min price like this:`quote.estimatedPrice * 0.99`

jerryafr, line 121:
```suggestion
    estimatedPrice: getPrice(depositAmount, srcAsset, egressAmount, destAsset),
```
not holding a strong opinion here, I just feel like the user doesn't care about ingress/egress fees. they just care about how much they sent and how much they got and that would be the effective exchange rate aka price for them

jerryafr (approved):

packages/swap/src/utils/getPoolQuote.ts
jerryafr: ah I see, so you're doing this to make it aligned with fok params. that's reasonable

packages/swap/src/routes/__tests__/__snapshots__/quote.test.ts.snap
```diff
@@ -4,7 +4,7 @@ exports[`server GET /quote doesn't include boost information when disabled 1`] =
 {
   "egressAmount": "999999999999975000",
   "estimatedDurationSeconds": 1830,
-  "estimatedPrice": "0.999999999999975",
+  "estimatedPrice": "1.001001001001001001",
```
niklasnatter, line 7: makes sense that `estimatedPrice` is higher because we dont include ingress and egress fee now

packages/swap/src/routes/__tests__/__snapshots__/quote.test.ts.snap
```diff
@@ -33,7 +33,7 @@ exports[`server GET /quote doesn't include boost information when disabled 1`] =
         "chain": "Bitcoin",
       },
       "fee": {
-        "amount": "200000",
+        "amount": "199800",
```
niklasnatter, line 36: makes sense that pool fee is lower because we deduct ingress fee before calculating it

packages/swap/src/utils/getPoolQuote.ts
niklasnatter: if you calculate `quote.estimatedPrice * 0.99` right now, you will accept a slightly higher slippage than 1% (or a significantly higher slippage if the ingress/egress fees are high relative to the quoted amount)

acdibble (approved):

packages/swap/src/utils/getPoolQuote.ts
niklasnatter: maybe we should use a more explicit name like `estimatedAmmPrice`/`estimatedFillPrice` and deprecate this one to make it more explicit

PR #986, opened 2024-10-02
chore(swap): include last update timestamp to responses [WEB-1419]

GabrielBuragev:


packages/swap/src/utils/intercept.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export const stalenessCheck = async (_: Request, res: Response, next: NextFunction) => {
+  const lastUpdateTimestamp = await getLastChainTrackingUpdateTimestamp();
+
+  if (lastUpdateTimestamp) {
+    res.header('X-Last-Update', lastUpdateTimestamp.toUTCString());
+    // If we didn't update the chain tracking table in the last minute, we consider the data stale
+    if (lastUpdateTimestamp.getTime() < Date.now() - 1000 * 60) {
+      res.header('X-Status', 'stale');
```
niklasnatter, line 25: `X-Status` is a bit generic, do we expect integrators to use this? then it might make sense to be more explicit, otherwise no strong opinion

niklasnatter (approved):

packages/swap/src/utils/intercept.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  return latestChainTracking.updatedAt;
+};
+
+export const stalenessCheck = async (_: Request, res: Response, next: NextFunction) => {
+  const lastUpdateTimestamp = await getLastChainTrackingUpdateTimestamp();
+
+  if (lastUpdateTimestamp) {
+    res.header('X-Last-Update', lastUpdateTimestamp.toUTCString());
```
GabrielBuragev, line 22: should we keep or remove this? I see no harm in having it, more visibility.

packages/swap/src/routes/__tests__/__snapshots__/swap.test.ts.snap
```diff
@@ -58,6 +58,7 @@ exports[`server GET /swaps/:id retrieves a swap in FAILED status (egress ignored
   "ignoredEgressAmount": "0",
   "intermediateAmount": "4506169140",
   "isDepositChannelExpired": false,
+  "lastStatechainUpdateAt": "2024-10-02T10:34:08.907Z",
```
niklasnatter, line 61: think there is a `valueOf()` missing as this should be a number

packages/swap/src/utils/intercept.ts
niklasnatter: fine for me to keep it but would call it `X-Last-Statechain-Update` then to keep the name consistent

packages/swap/src/routes/v2/swap.ts
```diff
@@ -239,6 +240,7 @@ router.get(
             swapDepositChannel?.failedBoosts.at(0)?.failedAtBlockIndex ?? undefined,
         },
       }),
+      lastStatechainUpdateAt: await getLastChainTrackingUpdateTimestamp(),
```
jerryafr, line 243: why do we need to add it here in addition to the headers?

packages/swap/src/utils/intercept.ts
acdibble: should be `X-CF-${string`

acdibble: oh no cloudflare

acdibble: `X-Flip-${string}`?

packages/swap/src/utils/intercept.ts
```diff
@@ -0,0 +1,26 @@
...
+  });
+
+  if (!latestChainTracking) {
+    return undefined;
+  }
+
+  return latestChainTracking.updatedAt;
+};
+
+export const stalenessCheck = async (_: Request, res: Response, next: NextFunction) => {
```
acdibble, line 18: need to use `asyncHandler` or try/catch. if one of these promises rejects, the rejected promise isn't handled by express >:(

packages/swap/src/routes/v2/swap.ts
jerryafr: missed the conversation in discord. makes sense. but is it possible that the sdk reads it from the header and put it in the final object?

packages/swap/src/utils/intercept.ts
GabrielBuragev: `X-Chainflip-...` ?

packages/swap/src/routes/v2/swap.ts
GabrielBuragev: Its a good idea but it comes with 2 disadvantages:
- Users calling the API endpoints directly (without using the SDK) will only have the headers to utilize
- Implicit knowledge required in the sdk methods to understand why we are checking for the headers there

packages/swap/src/utils/intercept.ts
GabrielBuragev: uf, good point!

packages/swap/src/utils/intercept.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  return latestChainTracking.updatedAt;
+};
+
+export const stalenessCheck = asyncHandler(
+  async (_: Request, res: Response, next: NextFunction) => {
+    const lastUpdateTimestamp = await getLastChainTrackingUpdateTimestamp();
+
+    if (lastUpdateTimestamp) {
+      res.header('X-Chainflip-Last-Statechain-Update', lastUpdateTimestamp.toUTCString());
```
niklasnatter, line 24: can we adjust the `CF_SDK_VERSION_HEADERS` too to follow this prefix?

niklasnatter: we dont use it atm so its fine to change it 🙂

GabrielBuragev: Sure

packages/swap/src/server.ts
```diff
@@ -27,8 +28,8 @@ const app = express().use(cors());
 const server = createServer(app);
 const io = new Server(server).use(authenticate);
 
-app.use('/swaps', express.json(), swap);
-app.use('/v2/swaps', express.json(), swapV2);
+app.use('/swaps', stalenessCheck, express.json(), swap);
```
niklasnatter, line 31:
```suggestion
app.use('/swaps', lastUpdateHeader, express.json(), swap);
```

niklasnatter (approved):

jerryafr (approved):

PR #989, opened 2024-10-02
fix(swap): Return witnessedAt for boosted swaps in getStatusV2

niklasnatter:
Boosted swaps do not have a `witnessedAt` at the moment even when they are completed: https://chainflip-swap.chainflip.io/v2/swaps/4642972-Bitcoin-11598


packages/swap/prisma/schema.prisma
```diff
@@ -139,6 +139,7 @@ model SwapRequest {
   srcAddress                   String?
   destAddress                  String?
   swapRequestedAt              DateTime
+  swapRequestedBlockIndex      String?
```
niklasnatter, line 142: we use this as `witnessedAt` for vault swaps

packages/swap/src/routes/v2/utils.ts
```diff
@@ -176,8 +176,15 @@ export const getDepositInfo = (
       witnessedAt:
         swapRequest?.originType === 'VAULT'
           ? swapRequest.swapRequestedAt.valueOf()
-          : swapRequest?.depositReceivedAt?.valueOf(),
-      witnessedBlockIndex: swapRequest?.depositReceivedBlockIndex ?? undefined,
+          : swapRequest?.depositBoostedAt?.valueOf() ??
+            swapRequest?.depositReceivedAt?.valueOf() ??
```
niklasnatter, line 180: a boosted swap does not have a `depositReceivedAt`

acdibble: probably easier to just check:

```ts
swapRequest?.depositBoostedBlockIndex ??
swapRequest?.depositReceivedBlockIndex ??
swapRequest?.swapRequestedBlockIndex
```

packages/swap/src/routes/v2/utils.ts
```diff
@@ -176,8 +176,15 @@ export const getDepositInfo = (
       witnessedAt:
         swapRequest?.originType === 'VAULT'
```
acdibble, line 177: same here, we can probably just remove this condition, especially considering the current behavior of swaps being requested in the same block as they are witnessed

niklasnatter: good with be too - updated

packages/swap/src/routes/v2/utils.ts
```diff
@@ -174,10 +174,15 @@ export const getDepositInfo = (
       txRef: depositTransactionRef,
       txConfirmations: pendingDeposit?.transactionConfirmations,
       witnessedAt:
-        swapRequest?.originType === 'VAULT'
-          ? swapRequest.swapRequestedAt.valueOf()
-          : swapRequest?.depositReceivedAt?.valueOf(),
-      witnessedBlockIndex: swapRequest?.depositReceivedBlockIndex ?? undefined,
+        swapRequest?.depositBoostedAt?.valueOf() ??
+        swapRequest?.depositReceivedAt?.valueOf() ??
+        swapRequest?.swapRequestedAt?.valueOf() ??
```
jerryafr, line 179: wouldn't this make a swap that's just waiting for deposit to return a wrong value?

niklasnatter: maybe i missunderstand the question, but if there is no deposit yet there should be no `swapRequest`

jerryafr: yes, so we're returning the swapRequest value as witnessed block even though there is no deposit yet, is that correct?

acdibble: a swap request is only created after a deposit is finalized or boosted

packages/swap/prisma/migrations/20241002133913_add_swap_scheduled_event/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "public"."SwapRequest" ADD COLUMN     "swapRequestedBlockIndex" TEXT;
```
acdibble, line 2: no backfill?

packages/swap/src/routes/v2/utils.ts
jerryafr: ah my bad, thanks

jerryafr (approved):

PR #991, opened 2024-10-04
fix(swap): Store transaction ref when deposit is boosted [WEB-1512]

niklasnatter:


acdibble (approved):

packages/swap/src/event-handlers/v140/__tests__/__snapshots__/depositBoosted.test.ts.snap
```diff
@@ -12,7 +12,7 @@ exports[`depositBoosted updates the values for an existing swap 1`] = `
   "depositBoostedBlockIndex": "120-7",
   "depositReceivedAt": null,
   "depositReceivedBlockIndex": null,
-  "depositTransactionRef": null,
+  "depositTransactionRef": "c7323b09ad750b311cdee3f1555c8ea9db52a7343a8d597474aa6c860f626b62",
```
niklasnatter, line 15: no need for a new test 🙌

PR #999, opened 2024-10-07
refactor(swap): add quoter logging

acdibble:


packages/swap/src/quoting/Quoter.ts
```diff
@@ -88,10 +90,18 @@ export default class Quoter {
...
+          ...result.data,
           marketMaker: socket.data.marketMaker,
         });
 
+        if (!this.inflightRequests.has(result.data.request_id)) {
+          logger.warn('received quote for unknown request', {
+            ...result.data,
+            marketMaker: socket.data.marketMaker,
+          });
+          return;
```
jerryafr, line 102: maybe we can clear that key from the Set here

jerryafr (approved):

PR #1001, opened 2024-10-08
fix(shared): fix dot validation

acdibble:


niklasnatter (approved):

packages/shared/src/broker.ts
```diff
@@ -82,12 +89,6 @@ const validateRequest = (network: ChainflipNetwork, params: unknown) => {
     .superRefine(ensureDcaWithFok)
     .parse(params);
 
-  if (parsed.destAsset.chain === 'Polkadot') {
```
niklasnatter, line 85: we sure that we can remove this? 😬

acdibble: yeas, a new case has been added to the union where we transform the ss58 address into hex

jerryafr (approved):

packages/shared/src/parsers.ts
```diff
@@ -91,14 +93,19 @@ export const ethereumAddress = hexString.refine(
...
+  (address) => ss58.decode(address).ss58Format === chainflipSS58Prefix,
   (address) => ({ message: `${address} is not a valid Chainflip address` }),
 );
 
 export const solanaAddress = string.refine(isValidSolanaAddress, (address) => ({
   message: `${address} is not a valid Solana address`,
 }));
 
+export const polkadotAddress = string.refine(
+  (address) => ss58.decode(address),
```
zoheb391, line 105: still need to limit length to 66

acdibble: `decode` checks that the address is 32 bytes. `hexString` doesn't have any limitation, which means that we don't check that EVM addresses either, i can add it for EVM too

zoheb391: ah didn't realise decode checked it 👍

zoheb391 (approved):

GabrielBuragev (approved):

PR #1004, opened 2024-10-08
feat: generate extrinsic params [WEB-1499]

acdibble:


packages/shared/src/broker.ts
```diff
@@ -164,3 +172,95 @@ export async function requestSwapDepositAddress(
...
+  swapRequest: NewSwapRequest,
+  chainflipNetwork: ChainflipNetwork,
+): SwappingRequestSwapDepositAddressWithAffiliates => {
+  const srcAsset = getInternalAsset({ asset: swapRequest.srcAsset, chain: swapRequest.srcChain });
+  const destAsset = getInternalAsset({
+    asset: swapRequest.destAsset,
+    chain: swapRequest.destChain,
+  });
+
+  const ccmParams = transformedCcmParamsSchema('0x')
```
niklasnatter, line 234: what does the default value do here?

packages/shared/src/broker.ts
```diff
@@ -50,21 +58,32 @@ const paramOrder = [
...
 
-const getAddressSchema = (network: ChainflipNetwork) =>
-  z.union([
-    numericString,
-    hexString,
-    btcAddress(network),
-    solanaAddress,
-    polkadotAddress.transform(ss58.toPublicKey),
-  ]);
+const getFokSchema = <Z extends z.ZodTypeAny>(addressSchema: Z) =>
```
niklasnatter, line 61:
```suggestion
const getTransformedFokSchema = <Z extends z.ZodTypeAny>(addressSchema: Z) =>
```

niklasnatter (approved):

packages/shared/src/broker.ts
acdibble: Currently we submit the ccm params to the broker without `cfParameters` I think. I wasn't sure what happens if you add the empty value (`0x`), but it's required for when you submit the extrinsic

PR #1007, opened 2024-10-09
fix(swap): use gas asset for ccm egress fee [WEB-1424]

GabrielBuragev:
The egress fees were stored using the wrong asset.

### DIFF
--- a/packages/sdk/src/swap/assets.ts
+++ b/packages/sdk/src/swap/assets.ts
@@ -12,7 +12,7 @@ import type { AssetData } from './types';
 
 const isGasAsset = (asset: InternalAsset) => {
   const { chain } = assetConstants[asset];
-  return assetConstants[asset].asset === chainConstants[chain].gasAsset;
+  return asset === chainConstants[chain].gasAsset;
 };
 
 export const getAssetData = (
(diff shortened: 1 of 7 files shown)


packages/swap/src/routes/swap.ts
```diff
@@ -145,7 +145,7 @@ router.get(
       pendingDeposit?.transactionHash ??
       failedSwap?.depositTransactionRef ??
       undefined;
-
+    console.log(swapRequest?.fees, swap?.fees);
```
zoheb391, line 148: missed

GabrielBuragev: woops

zoheb391 (approved):
can we add a test that checks this

jerryafr (approved):

acdibble (approved):

packages/sdk/src/swap/assets.ts
```diff
@@ -12,7 +12,7 @@ import type { AssetData } from './types';
 
 const isGasAsset = (asset: InternalAsset) => {
   const { chain } = assetConstants[asset];
-  return assetConstants[asset].asset === chainConstants[chain].gasAsset;
+  return asset === chainConstants[chain].gasAsset;
```
niklasnatter, line 15: im not sure if this is right - will this work for `ArbEth`?

niklasnatter: oh nevermind, just saw that you changed the type of `gasAsset`

PR #1008, opened 2024-10-09
fix(swap): ignore gas swaps for single asset ccm swaps [WEB-1424]

GabrielBuragev:


packages/swap/src/routes/v2/swap.ts
```diff
@@ -151,7 +155,6 @@ router.get(
       getRequiredBlockConfirmations(internalSrcAsset),
     ]);
 
-    const showCcm = Boolean(swapDepositChannel?.ccmGasBudget || swapDepositChannel?.ccmMessage);
```
GabrielBuragev, line 154: ccm params were only returned for deposit channel ccm swaps but not for SC ccm

packages/swap/src/routes/v2/swap.ts
```diff
@@ -72,37 +74,39 @@ router.get(
...
-        isDca: Boolean(
-          swapDepositChannel?.chunkIntervalBlocks && swapDepositChannel.chunkIntervalBlocks > 1,
-        ),
-        fees: [] as SwapFee[],
-      },
-    );
+    const swaps = swapRequest?.swaps.filter((swap) => swap.type !== 'GAS');
+
+    const rolledSwaps = swaps?.length
+      ? swaps.reduce(
```
acdibble, line 80:
```suggestion
    const rolledSwaps = swaps?.reduce(
```

jerryafr (approved):

packages/swap/src/routes/v2/swap.ts
GabrielBuragev: we don't want to return the 0s from the default value in the status response (check line 203). At least that felt most natural to me.

acdibble (approved):

packages/swap/src/routes/v2/swap.ts
GabrielBuragev: We just had an ingress -> gas swap (for the egress fee) -> egress . No principal or other swaps done.

acdibble: yeah but the optional chain will have the same result as this ternary

GabrielBuragev: hmm, not when the array of swaps is empty though? rolledSwaps will be initialized with the default value of the reducer

PR #1032, opened 2024-10-22
feat(swap): Allow to store quote when creating deposit channel through backend [WEB-1557]

niklasnatter:


packages/swap/prisma/schema.prisma
```diff
@@ -417,23 +418,19 @@ model StateChainError {
...
+  id                         Int                @id @default(autoincrement())
+  swapDepositChannel         SwapDepositChannel @relation(fields: [swapDepositChannelId], references: [id])
+  swapDepositChannelId       BigInt             @unique
+  srcAsset                   InternalAsset
+  destAsset                  InternalAsset
+  depositAmount              Decimal            @db.Decimal(30, 0)
+  intermediateAmount         Decimal?           @db.Decimal(30, 0)
+  egressAmount               Decimal            @db.Decimal(30, 0)
+  estimatedPrice             Decimal
+  executedPrice              Decimal?
```
niklasnatter, line 431: will set this in an upcoming pr

packages/shared/src/schemas.ts
```diff
@@ -120,6 +120,13 @@ export const openSwapDepositChannelSchema = z
     srcAddress: z.string().optional(),
     fillOrKillParams: fillOrKillParams.optional(),
     dcaParams: dcaParams.optional(),
+    quote: z
```
niklasnatter, line 123: will pass this from the sdk in an upcoming pr

packages/swap/prisma/schema.prisma
```diff
@@ -417,23 +418,19 @@ model StateChainError {
...
-  quoterFees         Json?
-  poolIntermediate   Decimal?      @db.Decimal(30, 0)
-  poolOutput         Decimal       @db.Decimal(30, 0)
-  poolDuration       Int?
-  poolFees           Json?
-  createdAt          DateTime      @default(now())
-  version            Int
-
-  @@index([srcAsset, destAsset])
+model Quote {
```
zoheb391, line 421: wondering if it would be useful to know if it was a boost/dca quote.

zoheb391: without checking the channel

zoheb391 (approved):

packages/swap/prisma/schema.prisma
niklasnatter: yeah true, might be convenient - will add two booleans

packages/swap/prisma/schema.prisma
jerryafr: what happens if a channel has been reused? we'd always store the latest executedPrice?

niklasnatter: im think ill store only the first one for now. if its already set, i wont overwrite it

jerryafr: I think we can just store the quote here and a link to the deposit channel already enables us to find the executed swaps. then we can generate the report that we want based on the 3 tables (quote, swapDepositChannel and swapRequest)

jerryafr (approved):

packages/swap/prisma/schema.prisma
niklasnatter: yeah thats true, but its less convenient. dont have a strong opinion though - im also fine with keeping this table to a minimum

packages/swap/prisma/schema.prisma
niklasnatter: will do this in a separate pr because i already built on this

PR #1033, opened 2024-10-22
feat(swap): Set executedPrice to quote when swap request completes [WEB-1557]

niklasnatter:


acdibble (approved):

zoheb391 (approved):

packages/swap/prisma/schema.prisma
```diff
@@ -419,19 +421,21 @@ model StateChainError {
...
-  egressAmount               Decimal            @db.Decimal(30, 0)
-  estimatedPrice             Decimal
-  executedPrice              Decimal?
-  executedSlippagePercentage Decimal?
+  depositAmount           Decimal            @db.Decimal(30, 0)
+  intermediateAmount      Decimal?           @db.Decimal(30, 0)
+  egressAmount            Decimal            @db.Decimal(30, 0)
+  estimatedPrice          Decimal
+  executedPrice           Decimal?
+  executedSlippagePercent Decimal?
```
jerryafr, line 439: since we're storing all the info needed for reporting here, maybe we can also store `quotedAt` and `executedAt` timestamps to know if someone used the channel late

jerryafr (approved):

packages/swap/prisma/schema.prisma
niklasnatter: thats a good idea! unfortuately we dont have a `quotedAt` at the moment because there is no timestamp on the quote, but ill think about how to do this

PR #1034, opened 2024-10-22
feat(swap): Add isDca and isBoost flags to stored quotes [WEB-1557]

niklasnatter:


packages/swap/prisma/schema.prisma
```diff
@@ -424,6 +424,8 @@ model Quote {
   swapDepositChannelId       BigInt             @unique
   srcAsset                   InternalAsset
   destAsset                  InternalAsset
+  isBoost                    Boolean
```
niklasnatter, line 427: i think there are two options
- duplicate some information and make this table convenient to use on its own
- keep the table to a minimum and join the channel + swap request for analysis
  - this would mean removing duplicated properties like `srcAsset`, `destAsset`, `isBoost`, `isDca`, `executedPrice` and `executedSlippagePercentage`

niklasnatter: i went with the first option because it seemed convenient for a quick analysis, but its not a strong opinion and im also fine with the other

acdibble (approved):

packages/swap/prisma/schema.prisma
jerryafr: if we're going with the first option, maybe it's better to store the boostFeeBps and DcaParams as well 🤔

niklasnatter: will store `boostFeeBps` and `numberOfChunks` then

GabrielBuragev (approved):

PR #1036, opened 2024-10-23
feat(sdk): Pass quote to swap backend when opening deposit channel [WEB-1557]

niklasnatter:


packages/sdk/src/swap/sdk.ts
```diff
@@ -494,7 +494,7 @@ export class SwapSDK {
         channelOpeningFee: result.channelOpeningFee,
       };
     } else {
-      response = await this.trpc.openSwapDepositChannel.mutate(depositAddressRequest);
+      response = await this.trpc.openSwapDepositChannel.mutate({ ...depositAddressRequest, quote });
```
niklasnatter, line 497: decided to pass the full quote bc we wont need an sdk update if we want to store additional data like quoted fees in the future

zoheb391, line 497: should we save a quote if the broker is defined too?

niklasnatter: i thought about it, but im not sure. it would be a separate request after the channel was created to collect the data, this will feel a bit like "phoning home" for integrators

zoheb391 (approved):

packages/sdk/src/swap/sdk.ts
acdibble, line 497: we don't use the function yet in the swapping app, we should add it to the v1 method as well

niklasnatter: the v1 function does not accept a quote at the moment. will change the swapping app to use v2 once this is merged 🙂

niklasnatter: the alternative would be changing the v1 method to additionally accept a quote, but feels wrong as we already have a v2 method that does this

jerryafr (approved):

PR #1037, opened 2024-10-24
feat: handle 1.8 protocol changes

albert-llimos:
- Add EVM Vault swaps new features (Fok/Boost/DCA/affiliates) via `cfParameters` encoding.
- Rename `cfParameters` to `ccmAdditionalData` in the `ccmParams`.
- Enforce FoK for Vault swaps.

The base for this is the current Cli used in the bouncer (1.6.3) and I've renamed it to 1.8.0 as that's the release on the backend that should contain those changes.

This PR will need some cleanup by some product team member so it fits nicely with everything else - e.g. update the `requestSwapDepositAddressWithAffiliates` from `@chainflip/extrinsics`, update tests as needed etc..

~~Only TBD left is the format of the `brokerFees` parameters. Right now it's hardcoded to an empty array when encoding.~~ Done


dandanlen:
FWIW, I can see why this is necessary for now, but we should not have to do this (parameter construction) in the SDK. It should be exposed via rpc on the node / broker api. This way there is much less risk of breakage or incompatible encoding.

niklasnatter:
Thanks a lot - all the adjustments were merged in separate PRs and we updated the bouncer in https://github.com/chainflip-io/chainflip-backend/pull/5712

PR #1038, opened 2024-10-24
feat(swap,sdk): Split up estimated durations into deposit, swap and egress [WEB-1524]

niklasnatter:


packages/swap/src/utils/swap.ts
```diff
@@ -18,24 +18,26 @@ export const estimateSwapDuration = async ({
...
+  const egressInclusionDuration = chainConstants[destChain].blockTimeSeconds;
 
-  return (
-    ingressTransactionDuration + ingressWitnessDuration + egressTransactionDuration + swapDuration
-  );
+  return {
+    deposit: depositInclusionDuration + depositWitnessDuration,
+    swap: swapDuration,
+    egress: egressInclusionDuration,
+  };
```
acdibble, line 40: can probably return `{ estimatedDurationSeconds: number, { durations: { ... } }` and then we can just spread it into the response

acdibble (approved):

jerryafr (approved):

PR #1045, opened 2024-10-28
chore(sdk): Allow to pass srcAddress in requestDepositAddressV2

niklasnatter:


CumpsD:
What's the use case for this?

niklasnatter:
> What's the use case for this?

It is one of the ways to prevent the chainflip broker from facilitating swaps for illicit funds. It will not be used if you configure your own broker

acdibble (approved):

zoheb391 (approved):

PR #1048, opened 2024-10-29
fix: no more need to transcode refund params

dandanlen:
See https://linear.app/chainflip/issue/PRO-1709/

Not sure how the package versioning is supposed to work, i followed Albert's PR...


acdibble:
merged into #1037

PR #1053, opened 2024-10-30
fix(swap): expose recommended slippage tolerance from getQuote

sukhoviianastasia:
Linear: [Expose recommended slippage tolerance from getQuote](https://linear.app/chainflip/issue/WEB-1556/expose-recommended-slippage-tolerance-from-getquote)

### DIFF
--- a/packages/shared/src/consts.ts
+++ b/packages/shared/src/consts.ts
@@ -87,3 +87,5 @@ export const MIN_TICK = -887272;
 export const MAX_TICK = -MIN_TICK;
 
 export const FULL_TICK_RANGE = { start: MIN_TICK, end: MAX_TICK };
+
+export const RECOMMENDED_SLIPPAGE_TOLERANCE_PERCENTAGE = 2;
--- a/packages/shared/src/schemas.ts
+++ b/packages/shared/src/schemas.ts
@@ -171,6 +171,7 @@ interface BaseQuoteDetails {
   lowLiquidityWarning: boolean | undefined;
   estimatedDurationSeconds: number;
   estimatedPrice: string;
+  recommendedSlippageTolerancePercent: number; 
 }
 
 type WithBoostQuote<T> = Omit<T, 'boostQuote'> & BoostedQuoteDetails;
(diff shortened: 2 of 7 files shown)


packages/shared/src/consts.ts
```diff
@@ -87,3 +87,5 @@ export const MIN_TICK = -887272;
 export const MAX_TICK = -MIN_TICK;
 
 export const FULL_TICK_RANGE = { start: MIN_TICK, end: MAX_TICK };
+
+export const RECOMMENDED_SLIPPAGE_TOLERANCE_PERCENTAGE = 2;
```
jerryafr, line 91: maybe making this an ENV variable would be better

jerryafr (approved):
Looks good! 🙌 
left a recommendation comment

packages/shared/src/consts.ts
acdibble: we discussed it, don't think it brings much since we want to replace it quite soon with a dynamic value but @niklasnatter can weigh in

acdibble (approved):

packages/shared/src/consts.ts
CumpsD: Dynamic coming from quoter endpoint? (aka https://chainflip-swap.chainflip.io/v2/quote)

niklasnatter: fine to have a constant for now - we will replace it with something dynamic relatively soon

niklasnatter (approved):
looks good to me! the next step would be opening a pull request in the github.com/chainflip-io/chainflip-docs/ repo to include the field in our documentation

packages/shared/src/schemas.ts
```diff
@@ -171,6 +171,7 @@ interface BaseQuoteDetails {
   lowLiquidityWarning: boolean | undefined;
   estimatedDurationSeconds: number;
   estimatedPrice: string;
+  recommendedSlippageTolerancePercent: number;
```
niklasnatter, line 174: maybe we should already add this to the changelog, even if its a static value for now? @acdibble

sukhoviianastasia: Added to changelog

niklasnatter (approved):
🙌

PR #1056, opened 2024-10-30
feat(swap): Add executedAt to Quote [WEB-1557]

niklasnatter:


packages/swap/prisma/migrations/20241030134031_add_executed_at_to_quote/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "private"."Quote" ADD COLUMN     "executedAt" TIMESTAMP(3);
```
niklasnatter, line 2: no migration needed as this is not live anywhere yet 🙂

PR #1069, opened 2024-11-06
feat(swap): [WEB-1568] auto calculate recommended slippage

jerryafr:


packages/shared/src/consts.ts
```diff
@@ -87,5 +87,3 @@ export const MIN_TICK = -887272;
 export const MAX_TICK = -MIN_TICK;
```
niklasnatter, line 87: would be great to replace the `CHANGELOG` entry with this:
```
- `SwapSDK.prototype.getQuoteV2` returns a `recommendedSlippageTolerancePercent` 
  property now. The value is calculated based on current market conditions to prevent 
  refunds while protecting against big price movements. It can be passed into
  the `requestDepositAddressV2` method to set the slippage tolerance for a swap.
```

packages/swap/src/utils/getPoolQuote.ts
```diff
@@ -1,15 +1,83 @@
...
+  destAsset,
+  depositAmount,
+  boostFeeBps,
+}: {
+  srcAsset: InternalAsset;
+  destAsset: InternalAsset;
+  depositAmount: bigint;
+  boostFeeBps?: number;
+}) => {
+  const minSlippage = 0.1;
```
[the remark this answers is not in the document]
jerryafr: yeah I'm not 100% sure about the current formula, but let's see

packages/swap/src/utils/getPoolQuote.ts
```diff
@@ -104,7 +172,12 @@ export default async function getPoolQuote<T extends QuoteType>({
   return {
     intermediateAmount: intermediateAmount?.toString(),
     egressAmount: egressAmount.toString(),
-    recommendedSlippageTolerancePercent: RECOMMENDED_SLIPPAGE_TOLERANCE_PERCENTAGE,
+    recommendedSlippageTolerancePercent: await calculateRecommendedSlippage({
```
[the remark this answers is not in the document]
jerryafr: will do it in a separate PR, this one is already too many changes :D

packages/swap/src/utils/__tests__/autoSlippage.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    const result = await calculateRecommendedSlippage({
+      srcAsset: 'Btc',
+      destAsset: 'Eth',
+      intermediateAmount: 3000n,
+      egressAmount: 1n,
+      dcaChunks: 100,
+    });
+
+    expect(result).toEqual(5.67);
```
jerryafr, line 113: I think this is a usual case (the swap amount and available liquidity) that we've daily. >5% slippage feels not so good here :/
probably we need to revise how formula

niklasnatter: yeah i expect us to tweak this quite a bit until we are happy. in this case, the `depositTimeSlippage` is too high imo

packages/swap/src/utils/autoSlippage.ts
```diff
@@ -0,0 +1,89 @@
...
+  destAsset: InternalAsset;
+  intermediateAmount?: bigint;
+  egressAmount: bigint;
+  boostFeeBps?: number;
+  dcaChunks: number;
+}) => {
+  const minSlippage = 0.1;
+  const baseSlippage = 1;
+
+  const assetTo = srcAsset === 'Usdc' || destAsset === 'Usdc' ? destAsset : 'Usdc';
```
niklasnatter, line 68: i think i dont understand this yet - does it mean we will only check the liquidity for the first leg?

packages/swap/src/utils/__tests__/autoSlippage.test.ts
niklasnatter: do you think we should hide the `recommendedSlippageTolerancePercent` behind a query param for now to not scare integrators?

packages/swap/src/utils/autoSlippage.ts
jerryafr: yeah, we only check it for the first leg because it includes the deposit time
we can include the second leg inside the calculations and get the Min (or Max?) value
let's discuss it

packages/swap/src/utils/__tests__/autoSlippage.test.ts
jerryafr: yeah that's a very good idea!
just put it behind `autoSlippageEnabled` query param

niklasnatter (approved):

packages/swap/src/utils/autoSlippage.ts
```diff
@@ -0,0 +1,122 @@
...
+  const baseSlippage = 1;
+
+  const [timeSlippage, deployedLiquiditySlippage, undeployedLiquiditySlippage] = await Promise.all([
+    getDepositTimeAdjustment(srcAsset, isBoosted),
+    getLiquidityAdjustment(srcAsset, destAsset, amount * dcaChunks),
+    getUndeployedLiquidityAdjustment(destAsset, amount * dcaChunks),
+  ]);
+
+  return deployedLiquiditySlippage
+    .plus(ignoreDepositTime ? 0 : timeSlippage)
```
acdibble, line 68:
```suggestion
    ignoreDepositTime ? 0 : getDepositTimeAdjustment(srcAsset, isBoosted),
    getLiquidityAdjustment(srcAsset, destAsset, amount * dcaChunks),
    getUndeployedLiquidityAdjustment(destAsset, amount * dcaChunks),
  ]);

  return deployedLiquiditySlippage
    .plus(timeSlippage)
```

niklasnatter (approved):

packages/swap/src/utils/autoSlippage.ts
```diff
@@ -0,0 +1,122 @@
...
+  const calculatedSlippageLeg2 = await calculateSingleLegSlippage({
+    srcAsset: 'Usdc',
+    destAsset,
+    amount: egressAmount,
+    isBoosted: !!boostFeeBps,
+    dcaChunks: BigInt(dcaChunks),
+    ignoreDepositTime: true,
+  });
+
+  const avgSlippage = calculatedSlippageLeg1.plus(calculatedSlippageLeg2).div(2).toNumber();
```
acdibble, line 120: not sure average is the best but we can try and see

acdibble, line 120: can also do it all with bignumber:
```suggestion
  const avgSlippage = calculatedSlippageLeg1.plus(calculatedSlippageLeg2).div(2).times(4).decimalPlaces(0).div(4).toNumber();
```

acdibble (approved):

packages/swap/src/utils/autoSlippage.ts
jerryafr: agree, we'll see how we should adjust it with real numbers

PR #1076, opened 2024-11-11
feat(swap): store recommendedSlippageTolerancePercent in quote table

jerryafr:


packages/shared/src/schemas.ts
```diff
@@ -127,6 +127,7 @@ export const openSwapDepositChannelSchema = z
         intermediateAmount: z.string().optional(),
         egressAmount: z.string(),
         estimatedPrice: z.string(),
+        recommendedSlippageTolerancePercent: z.number().optional(),
```
niklasnatter, line 130: think this can be non-optional

packages/swap/prisma/schema.prisma
```diff
@@ -424,25 +424,26 @@ model StateChainError {
...
+  numberOfChunks                      Int?
+  expectedDepositAmount               Decimal            @db.Decimal(30, 0)
+  quotedIntermediateAmount            Decimal?           @db.Decimal(30, 0)
+  quotedEgressAmount                  Decimal            @db.Decimal(30, 0)
+  quotedPrice                         Decimal
+  slippageTolerancePercent            Decimal?
+  executedAt                          DateTime?
+  executedPrice                       Decimal?
+  executedSlippagePercent             Decimal?
+  recommendedSlippageTolerancePercent Decimal?
```
niklasnatter, line 446: think this can be non nullable (can default to 0 in the migration)

packages/swap/src/handlers/openSwapDepositChannel.ts
```diff
@@ -121,6 +121,7 @@ export default async function openSwapDepositChannel(
       quotedEgressAmount: input.quote.egressAmount,
       quotedPrice: input.quote.estimatedPrice,
       slippageTolerancePercent: getSlippageTolerancePercent(input),
+      recommendedSlippageTolerancePercent: input.quote.recommendedSlippageTolerancePercent,
```
niklasnatter, line 124: we also need to add a new property to `enabledFeatures` of the sdk class to request the auto slippage

niklasnatter: could call it `experimentalRecommendedSlippage`

niklasnatter: but can also do in a separate pr 🙂

niklasnatter (approved):

packages/shared/src/schemas.ts
jerryafr: I feel like since it's not mandatory for the main purpose of this endpoint, would be better to have it as optional. would make it more error proof.
what do you think?

niklasnatter: fine for me as well - not a strong opinion 🙂

packages/swap/src/handlers/openSwapDepositChannel.ts
jerryafr: added it in this PR, would you please check again? :)

acdibble (approved):

CHANGELOG.md
```diff
@@ -28,6 +28,17 @@ which it is deprecated.
   property now. The value is calculated based on current market conditions to prevent
   refunds while protecting against big price movements. It can be passed into
   the `requestDepositAddressV2` method to set the slippage tolerance for a swap.
+- Optional parameter to enable recommended slippage during SDK initialization. The default value
```
niklasnatter, line 31: think i would not document this for now and enable it by default (remove the option) once we are confident. but not a blocker for me

niklasnatter (approved):

PR #1078, opened 2024-11-14
chore(swap): Use maximum of liquidity adjustment for recommended slippage [WEB-1568]

niklasnatter:


packages/swap/prisma/migrations/20241114163530_reset_recommended_slippage_column/migration.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `numberOfChunks` on the `SwapRequest` table. All the data in the column will be lost.
+
+*/
+-- AlterTable
+UPDATE "private"."Quote" q SET "recommendedSlippageTolerancePercent" = NULL;
```
niklasnatter, line 8: reset column to not include old data in analysis

acdibble (approved):
the data has spoken

PR #1082, opened 2024-11-18
feat(swap): handle swap rate v3

zoheb391:
I've tested this against perseverance and did not find any discrepancies except for DCA with boost. But since boost will change in 1.7.2 i think we can ignore that for now. (there is an open PR to add boost to this RPC)

closes WEB-1640


jerryafr (approved):
🙌

packages/swap/src/routes/v2/quote.ts
```diff
@@ -256,16 +195,23 @@ export const generateQuotes = async ({
     throw quoteResult.reason;
   }
 
-  const dcaQuote = dcaQuoteResult.value;
   const quote = getFulfilledResult(quoteResult, null);
   const boostedQuote = getFulfilledResult(boostedQuoteResult, null);
+  const dcaQuote = getFulfilledResult(dcaQuoteResult, null) as DCAQuote | null;
```
acdibble, line 200: this can be a DCA quote or a regular quote depending on the number of chunks

packages/swap/src/routes/v2/quote.ts
```diff
@@ -185,68 +136,56 @@ export const generateQuotes = async ({
...
-        depositAmount: dcaQuoteParams.chunkSize + ingressFeeSurcharge,
-        quoteType: 'DCA',
-        dcaChunks: dcaQuoteParams.numberOfChunks,
+        dcaParams,
+      }),
+    dcaQuoteParams &&
+      estimatedBoostFeeBps &&
+      getPoolQuote({
+        ...dcaQuoteArgs,
+        boostFeeBps: estimatedBoostFeeBps,
```
acdibble, line 187: should probably check at this point that we have > 1 chunk and not request a quote if we don't

packages/swap/src/utils/getPoolQuote.ts
```diff
@@ -88,6 +84,7 @@ export default async function getPoolQuote<T extends QuoteType>({
   includedFees.push(
     buildFee(getInternalAsset(ingressFee), 'INGRESS', ingressFee.amount),
     buildFee('Usdc', 'NETWORK', networkFee.amount),
+    ...[brokerFee.amount > 0n && buildFee('Usdc', 'BROKER', brokerFee.amount)].filter(Boolean),
```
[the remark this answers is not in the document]
zoheb391: i did it this way so we can preserve the "right" order of fees being taken out in the response.

packages/swap/src/routes/v2/quote.ts
niklasnatter: could do the check when setting `dcaParams` initially

niklasnatter (approved):
the changes look good to me! would be nice to compare the results against mainnet just to be safe

packages/swap/src/routes/v2/quote.ts
zoheb391: if we check the number of chunks before hand this can only be a `DCAQuote` ?

acdibble (approved):

PR #1084, opened 2024-11-20
chore(swap): Store refund information on Quote [WEB-1557]

niklasnatter:


jerryafr (approved):

packages/swap/src/event-handlers/swapRequestCompleted.ts
```diff
@@ -35,29 +35,39 @@ export default async function swapRequestCompleted({
...
-        executedSlippagePercent,
-        swapRequestId: swapRequest.id,
-        swapRequestNativeId: swapRequest.nativeId,
-      },
-    });
+      await prisma.quote.update({
+        where: { id: channelQuote.id },
+        data: {
+          executedAt: new Date(block.timestamp),
+          executedPrice,
```
zoheb391, line 65: we don't care about executed price if it was partially executed?

zoheb391 (approved):

packages/swap/src/event-handlers/swapRequestCompleted.ts
```diff
@@ -35,29 +35,39 @@ export default async function swapRequestCompleted({
...
+      const quotedPrice = channelQuote.quotedPrice.toNumber();
+      const executedPrice = Number(
+        getSwapPrice(
+          swapRequest.srcAsset,
+          swapRequest.swapInputAmount.toFixed(),
+          swapRequest.destAsset,
+          totalSwapOutputAmount.toFixed(),
+        ),
+      );
+      const executedSlippagePercent = (100 * (quotedPrice - executedPrice)) / quotedPrice;
```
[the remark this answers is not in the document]
niklasnatter: we could, but i think we dont want to because it would distort the data for analysis in two ways: 
- if a swap partially refunded, its executed slippage was artificially limited (it refunded because the executed slippage would have exceeded the limit). you would need to exclude these swaps to not skew the averages, but thats very easy to forget
- the recommended slippage of the quote was for the quoted amount, but the executed slippage is about the swapped amount. in case of a refund these amounts wont be same, so analyzing the difference between recommended and executed would be misleading

packages/swap/src/event-handlers/swapRequestCompleted.ts
niklasnatter: i think it would be kinda misleading to store it for partially refunded swaps because its artifically limited by the tolerance of the swap - added more on this in https://github.com/chainflip-io/chainflip-sdk-monorepo/pull/1084#discussion_r1851890222

PR #1085, opened 2024-11-21
chore(swap): Decrease maximum recommended slippage [WEB-1568]

niklasnatter:


packages/swap/src/utils/autoSlippage.ts
```diff
@@ -81,7 +81,7 @@ export const calculateRecommendedSlippage = async ({
   // use different limits for flip swaps because chainflip is the primary market for the flip token
   // because of this, lps cannot easily source liquidity from other markets (cex, dex) when filling a flip swap
   const MIN_SLIPPAGE = srcAsset === 'Flip' || destAsset === 'Flip' ? 1.5 : 0.5;
-  const MAX_SLIPPAGE = srcAsset === 'Flip' || destAsset === 'Flip' ? 5 : 3;
+  const MAX_SLIPPAGE = srcAsset === 'Flip' || destAsset === 'Flip' ? 5 : 2.5;
```
acdibble, line 84: if we create a weights table we can just update this in the db

acdibble (approved):

packages/swap/src/utils/autoSlippage.ts
niklasnatter: yes you are right, it might be time to make this configurable in some way. lets talk about it in the office

PR #1104, opened 2024-12-03
feat(swap): Try to find solana tx ref when solana deposit is finalized [WEB-1677]

niklasnatter:


packages/swap/src/utils/solana.ts
```diff
@@ -0,0 +1,84 @@
...
+        ?.uiTokenAmount.amount ?? 0,
+    );
+
+    return postBalance - preBalance;
+  }
+
+  throw new Error(`Unsupported asset: ${asset}`);
+};
+
+export const tryFindSolanaDepositTxRef = async (
```
niklasnatter, line 58: we will need the same logic in the explorer, so i am going to move this into the toolkit next

packages/swap/src/utils/solana.ts
```diff
@@ -0,0 +1,84 @@
...
+    filteredSignatures.map((sig) => sig.signature),
+    { maxSupportedTransactionVersion: 0 },
+  );
+
+  for (const tx of txs.filter(isNotNull)) {
+    const txAmount = getDepositAmount(tx, depositAddress, asset);
+    const txAmountRatio = BigNumber(txAmount.toString()).div(amount.toString()).toNumber();
+
+    // witnessed amount can be slightly higher than transaction amount if deposit is witnessed together with spam transactions
+    if (txAmountRatio > 0.95) {
```
acdibble, line 78:
```suggestion
    const txAmountRatio = BigNumber(txAmount.toString()).div(amount.toString());

    // witnessed amount can be slightly higher than transaction amount if deposit is witnessed together with spam transactions
    if (txAmountRatio.gt(0.95)) {
```

acdibble (approved):

acdibble (approved):

jerryafr (approved):

PR #1119, opened 2024-12-09
fix(swap): use maintenance mode middleware

GabrielBuragev:


packages/swap/src/server.ts
```diff
@@ -60,8 +60,8 @@ app.get('/healthcheck', (req, res) => {
   res.status(200).send('OK');
 });
 
-app.use('/quote', quoteRouter(quoter));
-app.use('/v2/quote', quoteRouterV2(quoter));
+app.use('/quote', maintenanceMode, quoteRouter(quoter));
+app.use('/v2/quote', maintenanceMode, quoteRouterV2(quoter));
```
GabrielBuragev, line 64: for some reason we are only using `DISABLE_QUOTING` on these 2 endpoints. 
When maintenance mode is turned on, the `MAINTENANCE_MODE` env is set to true. That should be sufficient for these 2 endpoints. Right now we need to manually set the `DISABLE_QUOTING` env var via argocd if we wanna disable these 2 endpoints when in maintenance mode.

PR #1122, opened 2024-12-10
refactor(swap): handle error

acdibble:


jerryafr (approved):

zoheb391 (approved):

packages/swap/src/routes/common.ts
```diff
@@ -44,3 +47,26 @@ export const asyncHandler = <
...
+  if (err instanceof ServiceError) throw err;
+
+  const message = err instanceof Error ? err.message : 'unknown error (possibly no liquidity)';
+
+  if (message.includes('InsufficientLiquidity') || message.includes('-32603')) {
+    logger.info('insufficient liquidity received', info);
+    throw ServiceError.badRequest('insufficient liquidity for requested amount');
+  }
+
+  logger.error('error while collecting quotes:', err);
```
GabrielBuragev, line 69: can we print the `info` object here as well? It would be nice to see which quote request this error is related to

PR #1123, opened 2024-12-11
fix(swap): validate total liquidity for non dca quotes

jerryafr:


packages/swap/src/routes/v2/quote.ts
```diff
@@ -234,29 +234,37 @@ export const generateQuotes = async ({
...
+    if (dcaBoostedQuote) {
+      adjustDcaQuote({ dcaQuoteParams, dcaQuote: dcaBoostedQuote, originalDepositAmount: amount });
+    }
+
+    // Check liquidity for DCA
+    await validateLiquidity(dcaQuote.egressAmount, dcaQuote.intermediateAmount);
+  } else if (quote) {
+    await validateLiquidity(quote.egressAmount, quote.intermediateAmount);
+  } else {
+    throw ServiceError.internalError('could not generate quotes');
```
niklasnatter, line 267: i think we should always do this for all of the quotes, no?

```suggestion
  } 
  
  if (quote) {
    await validateLiquidity(quote.egressAmount, quote.intermediateAmount);
  }
```

PR #1126, opened 2024-12-11
fix(swap): Allow to query status by non-hex transaction ref

niklasnatter:


packages/swap/src/routes/v2/utils.ts
```diff
@@ -95,7 +94,7 @@ export const getLatestSwapForId = async (id: string) => {
       where: { nativeId: BigInt(id) },
       include: swapRequestInclude,
     });
-  } else if (txHashRegex.test(id)) {
```
niklasnatter, line 98: solana transaction signatures are base58 encoded. we could also adjust the regex for that

packages/swap/src/routes/v2/utils.ts
```diff
@@ -95,7 +94,7 @@ export const getLatestSwapForId = async (id: string) => {
...
+  } else {
     swapRequest = await prisma.swapRequest.findFirst({
       where: { depositTransactionRef: id },
```
niklasnatter, line 99: i think it would be nice to make this case insensitive

packages/swap/src/routes/v2/utils.ts
niklasnatter: oof forgot about polkadot

packages/swap/src/routes/v2/utils.ts
GabrielBuragev: totally, i saw a lot of inconsistency when copying hashes from different places (sometimes checksummed,  sometimes not)

packages/swap/src/routes/v2/utils.ts
niklasnatter: okay i thought about this and decided to not use a regex here for now. the regex would save us one db request if somebody requests a status for something that is not a valid ref. i think that is negligible if it happens by mistake. and if somebody wants to spam us intentionally, he probably is sophisticated enough to use something that looks like a ref

niklasnatter: the upside of not having the regex is that we dont need to think about adjusting it if something changes (and we also cannot fuck up the regex lol)

GabrielBuragev: Yeah, i think i would be fine with opening this fallback so we don't need to handle each case separately and possibly save us some time in the future as well!

GabrielBuragev (approved):

PR #1128, opened 2024-12-12
refactor(swap): validate tx hash input

acdibble:


packages/swap/src/routes/v2/utils.ts
```diff
@@ -94,17 +97,17 @@ export const getLatestSwapForId = async (id: string) => {
       where: { nativeId: BigInt(id) },
       include: swapRequestInclude,
     });
-  } else if (id) {
+  } else if (hexRegex.test(id) || base58Regex.test(id)) {
```
niklasnatter, line 100: also need to accept polkadot txref format

niklasnatter (approved):
looks good to me otherwise

PR #1129, opened 2024-12-12
fix(swap): disconnect duplicate sockets

acdibble:


packages/swap/src/quoting/Quoter.ts
```diff
@@ -126,19 +126,30 @@ export default class Quoter {
...
     private readonly io: QuotingServer,
     private createId: () => string = randomUUID,
   ) {
     io.on('connection', (socket) => {
-      logger.info(`market maker "${socket.data.marketMaker}" connected`);
+      logger.info('market maker connected', { accountId: socket.data.accountId });
+
+      if (this.connectedAccounts.has(socket.data.accountId)) {
+        logger.warn('duplicate market maker connection', { accountId: socket.data.accountId });
+        socket.disconnect(true);
```
niklasnatter, line 140: can we send the reason to the client if he is using v2?

niklasnatter (approved):

zoheb391 (approved):

acdibble:
closed in favor of #1246

PR #1139, opened 2024-12-30
fix(swap): increase max slot when searching sol deposits

GabrielBuragev:


packages/swap/src/event-handlers/networkDepositFinalised.ts
```diff
@@ -44,7 +44,7 @@ export const networkDepositFinalised = async ({ prisma, event, block }: EventHan
 
     // somehow the blockHeight on the event can be slightly lower than the slot of the actual deposit
     // https://linear.app/chainflip/issue/PRO-1893/look-into-solana-ingress-block-height
-    const maxDepositSlot = Number(blockHeight) + 30;
+    const maxDepositSlot = Number(blockHeight) + 50;
```
jerryafr, line 47: I think increasing this would also increase the chance for overlaps. (use a tx incorrectly for 2 swaps).
but I don't see any way around it. maybe nik can confirm this later

jerryafr (approved):

packages/swap/src/event-handlers/networkDepositFinalised.ts
GabrielBuragev: Uf yeah, you are right.

GabrielBuragev: Took this discussion internally and decided to proceed with the `50` blocks as a buffer.

PR #1145, opened 2025-01-09
feat(swap): Store deposit tx ref for solana vault swaps [WEB-1677]

niklasnatter:


packages/swap/src/event-handlers/__tests__/swapRequested.test.ts
```diff
@@ -249,12 +292,35 @@ describe(swapRequested, () => {
...
+
+    const request = await prisma.swapRequest.findFirstOrThrow();
+
+    expect(request).toMatchSnapshot({
+      id: expect.any(BigInt),
+    });
+  });
+
+  it('creates a new swap request and gets the solana tx ref (VAULT 180)', async () => {
+    env.SOLANA_RPC_HTTP_URL = 'https://rpc.devnet.solana.chainflip.io:443';
```
zoheb391, line 305: this env is already specified in swap's vitest config, so there no need to set it here.

zoheb391 (approved):

GabrielBuragev (approved):

packages/swap/src/event-handlers/__tests__/swapRequested.test.ts
niklasnatter: nice - removed it

PR #1147, opened 2025-01-10
fix(swap): allow quoting when ingress fee can not be esimated

GabrielBuragev:


packages/swap/src/routes/__tests__/v2/quote.test.ts
```diff
@@ -429,6 +408,96 @@ describe('server', () => {
...
+
+      const { body, status } = await request(server).get(`/v2/quote?${params.toString()}`);
+
+      expect(status).toBe(200);
+      expect(body).toMatchObject([
+        {
+          egressAmount: '1000000000000000000',
+          recommendedSlippageTolerancePercent: 1,
+          includedFees: [
+            { type: 'INGRESS', chain: 'Ethereum', asset: 'USDC', amount: '0' },
```
niklasnatter, line 479: might make sense to filter the 0 fees out

niklasnatter (approved):

packages/swap/src/routes/__tests__/v2/quote.test.ts
GabrielBuragev: that affects also egress side so i wasn't sure if i should do it now, or maybe in a follow up

packages/swap/src/routes/v2/quote.ts
```diff
@@ -82,11 +82,7 @@ export const validateQuoteQuery = async (query: Query) => {
     throw ServiceError.badRequest(amountResult.reason);
   }
 
-  const ingressFee = await getIngressFee(srcAsset);
-
-  if (ingressFee === null) {
-    throw ServiceError.internalError(`could not determine ingress fee for ${srcAsset}`);
-  }
+  const ingressFee = (await getIngressFee(srcAsset)) ?? 0n;
```
jerryafr, line 85: maybe having a warning in the logs stating that it was null and therefore we fall back to zero would be useful

jerryafr (approved):

PR #1153, opened 2025-01-13
feat(swap): Handle InsufficientBoostLiquidity events for vault swaps [WEB-1764]

niklasnatter:


packages/swap/src/event-handlers/v140/insufficientBoostLiquidity.ts
```diff
@@ -1,17 +1,24 @@
...
+  originType: rustEnum(['Vault', 'DepositChannel']).optional(),
 });
-
 export const insufficientBoostLiquidity = async ({ prisma, event, block }: EventHandlerArgs) => {
-  const { channelId, asset, amountAttempted } = insufficientBoostLiquiditySchema.parse(event.args);
+  const { channelId, asset, amountAttempted, originType } = insufficientBoostLiquiditySchema.parse(
+    event.args,
+  );
+
+  // no deposit channel to look up failed boosts for vault swaps
```
zoheb391, line 18: don't  we want to make the deposit channel for boosts optional then? so we can capture this, or is that coming later?

zoheb391 (approved):

packages/swap/src/event-handlers/v140/insufficientBoostLiquidity.ts
niklasnatter: you are right, this would make sense - will change!

PR #1156, opened 2025-01-14
feat(swap): Refactor SwapDepositChannelAffiliate to include submitter [WEB-1750]

niklasnatter:


packages/swap/src/event-handlers/__tests__/__snapshots__/swapDepositAddressReady.test.ts.snap
```diff
@@ -215,5 +222,6 @@ exports[`swapDepositAddressReady > inserts affiliates 1`] = `
   "srcAsset": "Eth",
   "srcChain": "Ethereum",
   "srcChainExpiryBlock": 256n,
+  "totalBrokerCommissionBps": 125,
```
niklasnatter, line 225: its the sum of submitter and all affiliates now

packages/swap/src/routes/swap.ts
```diff
@@ -201,7 +207,8 @@ router.get(
       failure: failureMode,
       failedAt: failedSwap?.failedAt,
       failedBlockIndex: failedSwap?.failedBlockIndex ?? undefined,
-      depositChannelAffiliateBrokers: affiliateBrokers,
+      depositChannelAffiliateBrokers:
+        swapDepositChannel && affiliates?.length ? affiliates : undefined,
```
niklasnatter, line 211: check `affiliates?.length` to match response of old implementation

packages/swap/src/event-handlers/__tests__/__snapshots__/swapDepositAddressReady.test.ts.snap
niklasnatter: (we dont expose this value anywhere so its fine to change)

acdibble (approved):

zoheb391 (approved):

zoheb391:
just checking: the ticket has action items to move fok and dca params outside deposit channel, but i don't see those changes here.

PR #1157, opened 2025-01-14
fix(swap): filter out empty quoter limit orders

GabrielBuragev:


niklasnatter (approved):
strange that this would cause errors! would be nice to add a test

GabrielBuragev:
> strange that this would cause errors! would be nice to add a test

Yeah, actually we figured out its not a bug on our end. The quotes coming from CFG are using v2 schema (different parsing logic) while he subscribes with v1 client version 😅 Will add this guard in the `getSwapRateV3` instead.

acdibble (approved):

PR #1158, opened 2025-01-15
feat(swap): handle 1.8 event changes

acdibble:


packages/swap/src/event-handlers/__tests__/swapRequested.test.ts
```diff
@@ -80,41 +80,6 @@ const vaultSolana180 = {
   },
 } as const;
 
-const depositChannel160 = {
```
niklasnatter, line 83: not sure if we have a 170 event in the tests

packages/swap/src/event-handlers/refundEgressScheduled.ts
```diff
@@ -1,7 +1,9 @@
-import { swappingRefundEgressScheduled } from '@chainflip/processor/160/swapping/refundEgressScheduled';
+import { swappingRefundEgressScheduled as schema160 } from '@chainflip/processor/160/swapping/refundEgressScheduled';
+import { swappingRefundEgressScheduled as schema180 } from '@chainflip/processor/180/swapping/refundEgressScheduled';
+import z from 'zod';
 import type { EventHandlerArgs } from '.';
 
-const eventArgs = swappingRefundEgressScheduled;
+const eventArgs = z.union([schema180, schema160]);
```
niklasnatter, line 6: think we need to handle the new event shape here. typescript wont complain because of the `egressFee.toString()`

packages/swap/src/event-handlers/swapDepositAddressReady.ts
```diff
@@ -1,9 +1,10 @@
-import { swappingSwapDepositAddressReady } from '@chainflip/processor/160/swapping/swapDepositAddressReady';
+import { swappingSwapDepositAddressReady as schema160 } from '@chainflip/processor/160/swapping/swapDepositAddressReady';
+import { swappingSwapDepositAddressReady as schema180 } from '@chainflip/processor/180/swapping/swapDepositAddressReady';
 import { z } from 'zod';
 import { calculateExpiryTime } from '../utils/function';
 import { EventHandlerArgs } from './index';
 
-const swapDepositAddressReadyArgs = swappingSwapDepositAddressReady;
+const swapDepositAddressReadyArgs = z.union([schema180, schema160]);
```
niklasnatter, line 7: forgot what changed here again 😅 but i think we dont need to handle?

niklasnatter (approved):
❤️

packages/swap/src/config/__mocks__/env.ts
acdibble, line 1: a lot of tests were failing and it turned out to be because `MAINTENANCE_MODE` was `true` in my `.env` file.

This ensures that the `.env` file is never read and new variables will be `undefined` if not explicitly mocked.

packages/swap/src/event-handlers/refundEgressScheduled.ts
acdibble: true

packages/swap/src/event-handlers/swapDepositAddressReady.ts
acdibble: an enum value was added or removed I think. might not have to handle it in the logic, but we don't want the parsing to fail

PR #1161, opened 2025-01-15
test: tweak vitest settings

acdibble:


packages/swap/setupFile.ts
```diff
@@ -3,7 +3,7 @@ import { beforeAll, vi } from 'vitest';
 vi.mock('./src/config/env');
 
 beforeAll(() => {
-  global.fetch = vi
```
acdibble, line 6: this starting throwing an error saying you can't assign to a readonly property

PR #1178, opened 2025-01-27
chore(swap): test vault swap deposit failed is picked up by sdk

zoheb391:
closes WEB-1691

### DIFF
--- a/packages/swap/src/routes/__tests__/v2/swap.test.ts
+++ b/packages/swap/src/routes/__tests__/v2/swap.test.ts
@@ -1,7 +1,9 @@
+import { ethereumIngressEgressDepositFailed } from '@chainflip/processor/180/ethereumIngressEgress/depositFailed';
 import * as ss58 from '@chainflip/utils/ss58';
 import { Server } from 'http';
 import request from 'supertest';
 import { vi, describe, it, beforeEach, afterEach, expect, beforeAll } from 'vitest';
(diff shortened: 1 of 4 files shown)


packages/swap/src/routes/__tests__/v2/swap.test.ts
```diff
@@ -1048,7 +1138,7 @@ describe('server', () => {
       expect(body.deposit.failure).toMatchObject({
         failedAt: 552000,
         failedBlockIndex: '92-1',
-        mode: 'DEPOSIT_TOO_SMALL',
+        mode: 'DEPOSIT_IGNORED',
```
niklasnatter, line 1141: its a small breaking change, but i agree that its not very likely that somebody depends on this so no strong opinion

niklasnatter (approved):

packages/swap/src/routes/__tests__/v2/swap.test.ts
niklasnatter: could add it to the `CHANGELOG`

jerryafr (approved):

packages/swap/src/routes/__tests__/v2/swap.test.ts
zoheb391: apparently we don't check for it either 😅

PR #1189, opened 2025-02-05
feat(swap): add quoter beta mode [WEB-1945]

acdibble:


niklasnatter (approved):

packages/swap/src/quoting/Quoter.ts
```diff
@@ -207,11 +214,11 @@ export default class Quoter {
         this.inflightRequests.delete(request.request_id);
       };
 
-      sub = this.quotes$.subscribe(({ marketMaker, quote }) => {
+      sub = this.quotes$.subscribe(({ marketMaker, quote, beta }) => {
```
GabrielBuragev, line 217: are we allowing them to pass the `beta` state?

GabrielBuragev: ah nvm, just saw that we are setting the beta in the socket session based on the db value. All good

GabrielBuragev (approved):

GabrielBuragev (approved):
would be also nice to be able to provide an array of disabled LPs from the quoter via an env var (just like we do with the disabled assets) so we don't rely on a person with db write access if someone needs to intervene next time.

jerryafr (approved):

PR #1192, opened 2025-02-07
fix(swap): log 404s as warnings [WEB-1965]

acdibble:


zoheb391 (approved):

packages/swap/src/utils/disallowChannel.ts
```diff
@@ -31,7 +31,8 @@ const isTooRisky = async (address: string | undefined): Promise<boolean> => {
 
     return parsed.risk_score !== null && parsed.risk_score >= env.ELLIPTIC_RISK_SCORE_TOLERANCE;
   } catch (error) {
-    logger.error('failed to request risk score from elliptic', { error });
+    const level = isAxiosError(error) && error.status === 404 ? 'warn' : 'error';
+    logger[level]('failed to request risk score from elliptic', { error });
```
jerryafr, line 35:
```suggestion
    logger[level]('failed to request risk score from elliptic', { raw: error });
```
I think having the word `error` makes it still count as error on grafana

jerryafr (approved):

packages/swap/src/utils/disallowChannel.ts
acdibble: it now checks for `level=error`

PR #1201, opened 2025-02-10
chore(sdk): Check quote type in requestDepositAddressV2 and encodeVaultSwapData [WEB-1676]

niklasnatter:


packages/swap/src/utils/getPoolQuote.ts
```diff
@@ -142,5 +142,6 @@ export default async function getPoolQuote({
     srcAsset: getAssetAndChain(srcAsset),
     destAsset: getAssetAndChain(destAsset),
     depositAmount: depositAmount.toString(),
+    isVaultSwap,
```
acdibble, line 145: is the quoting different for vault swaps? is there any reason i can't take any quote and pass it to `encodeVaultSwapData`?

niklasnatter: yes, the vault swap quote does not have an ingress fee (or a different ingress fee for bitcoin)

niklasnatter: so passing a channel quote into `encodeVaultSwapData` would give you a better price than quote. which is nice i guess? but you also wont notice why

niklasnatter: the other direction (passing vault swap quote into channel) is def bad because you will get less money than you thought

acdibble (approved):

PR #1206, opened 2025-02-11
fix(swap): check number of chunks [WEB-1971]

tinkaa01:
Fix quoting when DCA is enabled.

We should check `numberOfChunks` to be less than 2**32, which is the maximum number of chunks the node can accept, since we have `u32` type for the `number_of_chunks` in our rpc method.

### DIFF
--- a/packages/swap/src/routes/v2/quote.ts
+++ b/packages/swap/src/routes/v2/quote.ts
@@ -18,6 +18,8 @@ import ServiceError from '../../utils/ServiceError';
 import { asyncHandler, handleQuotingError } from '../common';
 import { fallbackChains } from '../quote';
 
+export const MAX_NUMBER_OF_CHUNKS = 2n ** 32n - 1n;
+
 const router = express.Router();
 
 export const getDcaQuoteParams = async (asset: InternalAsset, amount: bigint) => {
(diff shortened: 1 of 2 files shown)


packages/swap/src/routes/v2/quote.ts
```diff
@@ -35,6 +37,11 @@ export const getDcaQuoteParams = async (asset: InternalAsset, amount: bigint) =>
   }
   const numberOfChunks = Math.ceil(Number(usdValue) / usdChunkSize);
 
+  if (numberOfChunks > MAX_NUMBER_OF_CHUNKS) {
```
acdibble, line 40: there is a much lower upper limit we can check:

```ts
const BLOCKS_PER_CHUNK = 2;
const SECONDS_PER_BLOCK = 6;
const MAX_DCA_DURATION_SECONDS = hoursToSeconds(24);
const MAX_NUMBER_OF_CHUNKS = MAX_DCA_DURATION_SECONDS / SECONDS_PER_BLOCK / BLOCKS_PER_CHUNK; // 7200
```

the protocol limits the total duration of a DCA swap to be at most 24 hours

tinkaa01: good to know, thank you! 🙌

packages/swap/src/routes/v2/quote.ts
```diff
@@ -35,6 +42,13 @@ export const getDcaQuoteParams = async (asset: InternalAsset, amount: bigint) =>
...
+    logger.error(
+      `number of chunks (${numberOfChunks}) is bigger than max (${MAX_NUMBER_OF_CHUNKS})`,
+    );
```
acdibble, line 48:
```suggestion
    logger.info('number of chunks is bigger than max', { numberOfChunks, MAX_NUMBER_OF_CHUNKS });
```

acdibble (approved):

PR #1209, opened 2025-02-12
fix(swap): Do not change parameter object in requestSwapDepositAddress

niklasnatter:


packages/shared/src/broker.ts
```diff
@@ -184,10 +184,12 @@ export async function requestSwapDepositAddress(
 
   /** @deprecated pass the chain in the srcAsset and destAsset object instead */
   if (request.srcChain && typeof request.srcAsset === 'string') {
-    request.srcAsset = { asset: request.srcAsset, chain: request.srcChain };
```
niklasnatter, line 187: the `request` is a parameter of this method. we pass the same object to `getSlippageTolerancePercent` here
https://github.com/chainflip-io/chainflip-sdk-monorepo/blob/069b56d4ce53f4b349ac001a8070db49cde3a5c9/packages/swap/src/handlers/openSwapDepositChannel.ts#L121

niklasnatter: to prevent this from happening anywhere, we set `request` to a new object instead of modifying the one that was passed as param

jerryafr (approved):

packages/shared/src/broker.ts
GabrielBuragev: Should we maybe completely avoid changing the request object here? We can move that responsibility outside of this method.
If we keep this apporach, then a test would be nice.

niklasnatter: its a backwards compatibility measure - we dont change it if you pass the non-deprecated params 🙂 but yeah we pass the deprecated params from `openSwapDepositChannel` at the moment

GabrielBuragev: I think we can still maintain BW without directly changing the request object but just create a local one. 
I understand that this is a convergence point and thats why we chose to overwrite the params here, but not the nicest by design as its hard to reason about.

PR #1214, opened 2025-02-17
feat: Support ccm_additional_data param [WEB-1889]

niklasnatter:


packages/swap/src/handlers/__tests__/__snapshots__/openSwapDepositChannel.test.ts.snap
```diff
@@ -281,6 +281,7 @@ exports[`openSwapDepositChannel > creates channel with ccmParams and stores it i
   [
     {
       "ccmParams": {
+        "additionalData": undefined,
```
niklasnatter, line 284: the output type of the `ccmParamsSchema` has a required `additionalData` key now because of the fallback

packages/shared/src/broker.ts
```diff
@@ -55,12 +55,15 @@ const transformedDcaParamsSchema = dcaParamsSchema.transform(
...
-  ccmParamsSchema.transform(({ message, gasBudget, cfParameters }) => ({
+const transformedCcmParamsSchema = ccmParamsSchema.transform(
+  ({ message, gasBudget, additionalData }) => ({
     message,
     gas_budget: gasBudget,
-    cf_parameters: cfParameters ?? defaultValue,
-  }));
+    ccm_additional_data: additionalData,
+    /** @deprecated DEPRECATED(1.8) we still need to pass cf_parameters until 1.8 is deployed to all networks */
+    cf_parameters: additionalData,
```
niklasnatter, line 64: tested against the backspin broker (1.8) and the persa broker (1.7). passing both keys for now seems to do the trick

acdibble (approved):

PR #1221, opened 2025-02-18
fix(sdk): return aborted ccm payload in the sdk[WEB-1865]

sukhoviianastasia:
`Signed transaction payload` returned on the swap id page should return these fields: `data`, `value`, `chainId` and `contract` in case if  broadcast was aborted. 
Added new field `transactionPayload ` to the object returned from `getEgressStatusFields` - it will return parsed payload when broadcast was aborted


packages/swap/src/routes/v2/utils.ts
```diff
@@ -312,6 +313,24 @@ export const getEgressStatusFields = async (
   const broadcast = egress?.broadcast;
   const ignoredEgress = ignoredEgresses?.find((e) => e.type === type);
   const failureState = await getEgressFailureState(ignoredEgress, broadcast, type);
+
+  let transactionPayload = broadcast?.transactionPayload;
+
+  if (broadcast?.abortedBlockIndex) {
+    const broadcastPayloadObj = JSON.parse(broadcast?.transactionPayload ?? '{}');
+    const broadcastPayloadObjParsed = cfChainsEvmTransaction.safeParse(broadcastPayloadObj);
+    if (broadcastPayloadObjParsed.success) {
```
sukhoviianastasia, line 322: do nothing in case `safeParse` is not successful, meaning payload is unchanged, not sure if it's ok 🤔

packages/swap/src/routes/v2/utils.ts
```diff
@@ -312,6 +313,24 @@ export const getEgressStatusFields = async (
   const broadcast = egress?.broadcast;
   const ignoredEgress = ignoredEgresses?.find((e) => e.type === type);
   const failureState = await getEgressFailureState(ignoredEgress, broadcast, type);
+
+  let transactionPayload = broadcast?.transactionPayload;
```
acdibble, line 317:
```suggestion
  let transactionPayload;
```

packages/swap/src/routes/v2/utils.ts
```diff
@@ -312,6 +313,24 @@ export const getEgressStatusFields = async (
...
+    const broadcastPayloadObjParsed = cfChainsEvmTransaction.safeParse(broadcastPayloadObj);
+    if (broadcastPayloadObjParsed.success) {
+      const { data, value, chainId, contract } = broadcastPayloadObjParsed.data;
+      const stringifiedFields = {
+        data,
+        value: value.toString(),
+        chainId: chainId.toString(),
+        contract,
+      };
+      transactionPayload = JSON.stringify(stringifiedFields);
```
acdibble, line 330:
```suggestion
      transactionPayload = {
        data,
        value: value.toString(),
        chainId: chainId.toString(),
        contract,
      };
```

packages/swap/src/routes/v2/utils.ts
acdibble: yeah i think it's the safest thing for now, we can add a `warn` log that we weren't able to parse the payload maybe

acdibble (approved):

GabrielBuragev (approved):

packages/sdk/src/swap/v2/types.ts
```diff
@@ -112,6 +114,7 @@ interface EgressFields {
   failure: Failure | undefined;
   failedAt: number | undefined;
   failedBlockIndex: string | undefined;
+  transactionPayload: z.infer<typeof cfChainsEvmTransaction> | undefined;
```
niklasnatter, line 117: i think this is not completely right, because we are calling `toString()` for some of the properties. but also its a bit hard for an integrator to understand what the type actually is if we define it this way. lets define it explicitly to make that easier:
```suggestion
  transactionPayload: {
    contract: string,
    ...
  }
```

packages/sdk/package.json
```diff
@@ -52,7 +52,8 @@
   },
   "dependencies": {
     "@chainflip/bitcoin": "^1.1.1",
-    "@chainflip/rpc": "^1.8.5",
+    "@chainflip/processor": "^1.8.4",
```
sukhoviianastasia, line 55: for some reason made me add these

acdibble: you can probably remove this now that there is an explicit type

niklasnatter (approved):

PR #1222, opened 2025-02-19
feat(swap): drop ccmDepositReceivedBlockIndex column [WEB-1866]

tinkaa01:


packages/shared/src/parsers.ts
```diff
@@ -126,6 +126,7 @@ export const assetAndChain = uncheckedAssetAndChain.refine((value): value is Ass
   isValidAssetAndChain(value as UncheckedAssetAndChain),
 );
 
+// TODO: remove this
 export const swapType = z.union([
```
tinkaa01, line 130: i think we don't use `swapType` parser at all, can we delete it?

niklasnatter: yes if we dont use it we can just remove

niklasnatter (approved):

PR #1227, opened 2025-02-21
chore(swap): Make encodeVaultSwapData a mutation

niklasnatter:


GabrielBuragev (approved):

packages/swap/src/server.ts
```diff
@@ -31,7 +31,7 @@ const appRouter = router({
     .mutation((v) => openSwapDepositChannel(v.input)),
   encodeVaultSwapData: publicProcedure
     .input(encodeVaultSwapDataSchema)
-    .query((v) => encodeVaultSwapData(v.input)),
+    .mutation((v) => encodeVaultSwapData(v.input)),
```
niklasnatter, line 34: while this is not really a mutation, registering it as one makes it consistent to the `openSwapDepositChannel` method and allows us to use the same configuration for stuff like rate limiting

PR #1229, opened 2025-02-21
fix(swap): Support parsing vault swaps without dca and ccm params from ingress-egress-tracker

niklasnatter:


packages/shared/src/node-apis/redis.ts
```diff
@@ -136,33 +136,38 @@ const vaultDepositSchema = (network: ChainflipNetwork) =>
             chunk_interval: number,
             number_of_chunks: number,
           })
+          .nullable()
```
niklasnatter, line 139: turns out `optional()` does not like it if the value of the key is `null`

PR #1230, opened 2025-02-21
fix: Include estimated durations in status for pending vault swaps

niklasnatter:


packages/swap/src/routes/v2/swap.ts
```diff
@@ -79,10 +79,6 @@ router.get(
         effectiveBoostFeeBps = 0;
       }
     }
-    const { srcAsset, destAsset } = {
```
niklasnatter, line 82: only used these variable for estimating the swap duration. also we already had `internalSrcAsset` and `internalDestAsset`

acdibble (approved):

PR #1232, opened 2025-02-22
refactor(swap): handle double patch version

acdibble:


packages/swap/src/utils/function.ts
```diff
@@ -93,10 +93,25 @@ const cachedGetSpecVersion = memoize(
   6_000,
 );
 
-export const isAfterSpecVersion = async (specVersion: number) => {
+// TODO(1.8): undo this commit, it assumes 4 digits will be a double patch and not double minor or major
```
GabrielBuragev, line 96: `double-digit` patch perhaps?

packages/swap/src/utils/__tests__/function.test.ts
```diff
@@ -0,0 +1,25 @@
...
+import * as rpc from '@/shared/rpc';
+import { isAtLeastSpecVersion } from '../function';
+
+vi.mock('@/shared/rpc');
+
+describe(isAtLeastSpecVersion, () => {
+  it('handles 1.7.10', async () => {
+    vi.mocked(rpc.getRuntimeVersion).mockResolvedValueOnce({ specVersion: 1710 } as any);
+
+    expect(await isAtLeastSpecVersion('1.8.0')).toBe(false);
```
GabrielBuragev, line 11: can we add a truthy test for the double-digit patch case?

GabrielBuragev (approved):

PR #1233, opened 2025-02-25
refactor(swap): add request id to logs

acdibble:
```json
{"component":"SWAP","error":{},"info":{"amount":"1","destAsset":"Usdc","srcAsset":"Eth","usdValue":"9800"},"level":"error","message":"error while collecting quotes:","reqId":"ccf291f8-5439-4c88-aaeb-8964fd6c143d","timestamp":"25-02-25 13:23:31"}
```

### DIFF
--- a/packages/swap/src/server.ts
+++ b/packages/swap/src/server.ts
@@ -23,7 +23,7 @@ import quoteRouterV2 from './routes/v2/quote';
 import swapV2 from './routes/v2/swap';
 import { publicProcedure, router } from './trpc';
 import { lastUpdateHeader } from './utils/intercept';
-import logger from './utils/logger';
+import logger, { logStorage } from './utils/logger';
 
 const appRouter = router({
   openSwapDepositChannel: publicProcedure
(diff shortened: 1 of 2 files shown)


GabrielBuragev (approved):

packages/swap/src/server.ts
```diff
@@ -58,7 +58,7 @@ app.use((req, res, next) => {
     });
   });
 
-  next();
+  logStorage.run(info.reqId, next);
```
jerryafr, line 61: can we do like
```
logStorage.run(info.reqId);
next();
```
so that it'd log in a non blocking way?

acdibble: This is non blocking. It create a storage around the call stack so that anywhere we call `logStorage.getStore()`, it returns the UUID

acdibble: This is basically the textbook example of when to use it https://nodejs.org/api/async_context.html#class-asynclocalstorage

PR #1234, opened 2025-02-25
fix(swap): Use hex encoded transaction id to find pending bitcoin vault swaps

niklasnatter:


jerryafr (approved):

packages/shared/src/node-apis/redis.ts
```diff
@@ -264,7 +265,11 @@ export default class RedisClient {
     const responses = await Promise.all(
       chainflipChains
         .filter((chain) => !vaultSwapDisabledChains.includes(chain))
-        .map((chain) => this.client.get(`vault_deposit:${chain}:${txId}`)),
+        .map((chain) => {
+          const redisTxId =
+            chain === 'Bitcoin' && isHex(`0x${txId}`) ? reverseBytes(`0x${txId}`) : txId;
```
acdibble, line 270:
```suggestion
            chain === 'Bitcoin' ? reverseBytes(`0x${txId}`) : txId;
```

acdibble (approved):

packages/shared/src/node-apis/redis.ts
niklasnatter: tried this first, but then `reverseBytes` throws if for a solana txref (base58 instead of hex) 😬

niklasnatter: we should probably filter the chains we query by the type of the txhash, but i did not want to change too much

PR #1235, opened 2025-02-26
feat: Add example scripts for triggering vault swaps

niklasnatter:


.github/workflows/_01_pre_checks.yml
```diff
@@ -46,13 +46,13 @@ jobs:
           done
 
       - name: Run prettier check 💅
-        run: pnpm --filter ${{ matrix.package }} prettier:check
+        run: pnpm --dir=packages/${{ matrix.package }} prettier:check
```
niklasnatter, line 49: this will fail now if one of the packages does not have the script

PR #1237, opened 2025-02-26
chore: Make test scripts consistent

niklasnatter:


packages/sdk/package.json
```diff
@@ -30,7 +30,7 @@
   ],
   "scripts": {
     "test": "vitest",
-    "test:ci": "CI=1 pnpm vitest run",
+    "test:ci": "CI=1 pnpm test",
```
niklasnatter, line 33: vitest will use "run" automatically if the `CI` env var is set

PR #1246, opened 2025-03-05
feat(swap): check LP balances [WEB-2042]

acdibble:


packages/swap/src/quoting/__tests__/BalanceTracker.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      .mockRejectedValueOnce(new Error())
+      .mockRejectedValueOnce(new Error())
+      .mockRejectedValueOnce(new Error())
+      .mockResolvedValueOnce([[lp1, mockInternalAssetMap()]]);
+    tracker['monitoredAccounts'].add(lp1);
+    expect(await tracker.getBalances()).toEqual(new Map());
+    expect(getLpBalances).toHaveBeenCalledTimes(5);
+  });
+
+  it('aborts and refetches if two lps are added around the same time', async () => {
```
niklasnatter, line 122: abort?

packages/swap/src/quoting/Quoter.ts
```diff
@@ -58,28 +60,44 @@ export type RpcLimitOrder = {
...
+): IteratorObject<RpcLimitOrder> {
+  if (!leg) return;
+
+  const sellAsset = getInternalAsset(leg.side === 'BUY' ? leg.base_asset : leg.quote_asset);
+
+  for (const { orders, balances } of quotes) {
+    const balance = balances?.[sellAsset];
+
+    for (const [tick, amount] of orders) {
+      if (balance === undefined || isBalanceWithinTolerance(balance, amount)) {
```
niklasnatter, line 82: would be nice to emit a `quote_error` to let the lp know if we ignore

packages/swap/src/quoting/Quoter.ts
```diff
@@ -58,28 +60,44 @@ export type RpcLimitOrder = {
...
+  return paddedBalance >= amount;
 };
 
+function* formatLimitOrders(
+  quotes: { orders: MarketMakerQuote['legs'][number]; balances?: InternalAssetMap<bigint> }[],
+  leg?: MarketMakerQuoteRequest<LegJson>['legs'][number],
+): IteratorObject<RpcLimitOrder> {
+  if (!leg) return;
+
+  const sellAsset = getInternalAsset(leg.side === 'BUY' ? leg.base_asset : leg.quote_asset);
```
niklasnatter, line 76: we are sure that `leg.side` is from the perspective or the swapper, right? 😅

packages/swap/src/quoting/Quoter.ts
```diff
@@ -58,28 +60,44 @@ export type RpcLimitOrder = {
...
+            sell_amount: hexEncodeNumber(amount),
+          },
+        };
+      }
+    }
+  }
+}
+
 type ClientVersion = '2';
+export type AccountId = string & { __brand: 'AccountId' };
```
niklasnatter, line 98: this is to prevent passing random strings right? might be abit confusing if you dont have the context

packages/swap/src/quoting/BalanceTracker.ts
```diff
@@ -0,0 +1,66 @@
...
+    this.refresh();
+  }
+
+  remove(account: AccountId) {
+    this.monitoredAccounts.delete(account);
+  }
+
+  async getBalances() {
+    if (this.lastUpdate < Date.now() - 10_000 && this.inflightCount === 0) {
+      await this.refresh();
```
jerryafr, line 32: maybe we can refresh here without await to keep it fast.
but if it's older than 30 seconds, we can await to ensure it's up to date

jerryafr (approved):
looks neat!

packages/swap/src/quoting/Quoter.ts
acdibble: yeah it's a pretty common pattern that we don't use anywhere yet but yeah it's to help keep track of this specific type of string throughout the logic

packages/swap/src/quoting/Quoter.ts
acdibble: yeah, you can see in the test that the USDC balance is checked when the source asset is BTC

packages/swap/src/quoting/__tests__/BalanceTracker.test.ts
acdibble: fixed

packages/swap/src/quoting/BalanceTracker.ts
acdibble: tried to refactor it to be a little more eager in it's fetching but couldn't come up with something i liked. for now we can see how much the quoting increases by and then adjust from there if necessary

packages/swap/src/quoting/Quoter.ts
acdibble: added a comment

PR #1253, opened 2025-03-10
chore(swap): Normalize pending vault swap property names [WEB-2028]

niklasnatter:


packages/swap/src/routes/v2/utils.ts
```diff
@@ -428,15 +410,34 @@ export const getSwapState = async (
   };
 };
 
+export const getBeneficiaries = (
```
niklasnatter, line 413: todo: add vault swap beneficiaries test

GabrielBuragev (approved):

PR #1256, opened 2025-03-13
chore(swap): Detect bitcoin vault swaps for mempool transactions [WEB-2028]

niklasnatter:


packages/swap/src/routes/v2/utils.ts
```diff
@@ -459,21 +459,18 @@ export const getDcaParams = (
...
+  const numberOfChunks =
+    swapRequest?.dcaNumberOfChunks ??
+    swapDepositChannel?.dcaNumberOfChunks ??
+    pendingVaultSwap?.dcaParams?.numberOfChunks;
+  const chunkIntervalBlocks =
+    swapRequest?.dcaChunkIntervalBlocks ??
+    swapDepositChannel?.dcaChunkIntervalBlocks ??
+    pendingVaultSwap?.dcaParams?.chunkInterval;
+
+  return numberOfChunks && numberOfChunks > 1 ? { numberOfChunks, chunkIntervalBlocks } : undefined;
```
niklasnatter, line 472: vault swaps always have the dca params encoded, even if `numberOfChunks` is 1

acdibble: ok this should be fixed in the bitcoin package, i have the logic wrong, wasn't sure what it would actually look like

GabrielBuragev (approved):

acdibble (approved):

PR #1260, opened 2025-03-17
chore: Update @chainflip/rpc package

niklasnatter:


packages/shared/src/broker.ts
```diff
@@ -206,15 +206,15 @@ export async function requestSwapDepositAddress(
...
+    'broker_request_swap_deposit_address',
     params.srcAsset,
     params.destAsset,
     params.destAddress,
     params.commissionBps,
     params.ccmParams,
     params.maxBoostFeeBps,
     params.affiliates,
-    params.fillOrKillParams,
+    params.fillOrKillParams!,
```
niklasnatter, line 217: wanted to keep this pr related to the package update. will property adjust types in a follow up together with validation + changelog

GabrielBuragev (approved):

PR #1262, opened 2025-03-18
feat: Support quoting for CCM swaps [WEB-2005]

niklasnatter:


packages/swap/src/utils/statechain.ts
```diff
@@ -29,15 +32,23 @@ export const getSwapRateV3 = async ({
...
+    ? {
+        number_of_chunks: _dcaParams.numberOfChunks,
+        chunk_interval: _dcaParams.chunkIntervalBlocks,
+      }
+    : undefined;
+  const ccmParams = _ccmParams
     ? {
-        number_of_chunks: dcaParams.numberOfChunks,
-        chunk_interval: dcaParams.chunkIntervalBlocks,
+        // TODO: remove cast once https://github.com/chainflip-io/chainflip-product-toolkit/pull/325 is merged
```
niklasnatter, line 49: will update package in a follow up bc the changed types will require some more changes

packages/sdk/src/swap/types.ts
```diff
@@ -45,23 +45,27 @@ export interface QuoteRequest extends ChainsAndAssets {
   brokerCommissionBps?: number;
   affiliateBrokers?: AffiliateBroker[];
   isVaultSwap?: boolean;
+  ccmParams?: {
+    gasBudget: string;
+    messageLengthBytes: number;
```
jerryafr, line 50: wdyt about getting the message: string here in the SDK instead of length in bytes to make it easier for the user?

niklasnatter: i thought about it, but the problem is that the message will depend on the quote for aggregation usecases. eg if you want to swap on the destination chain you will first get a quote to estimate how much tokens you will receive from Chainflip and then build the calldata for the smart contract call

packages/shared/src/schemas.ts
```diff
@@ -48,11 +50,26 @@ export const quoteQuerySchema = z
...
+      });
+
+      return z.NEVER;
+    }
+
+    const ccmParams =
+      args.ccmGasBudget !== undefined && args.ccmMessageLengthBytes !== undefined
+        ? { gasBudget: args.ccmGasBudget, messageLengthBytes: args.ccmMessageLengthBytes }
+        : undefined;
+
```
jerryafr, line 66: we would ignore ccm params if they provide the length and not the budget?

packages/sdk/src/swap/types.ts
niklasnatter: yes exactly. depending on how the receiving contract works you might already know the length because the params always have the same size, but otherwise it will be an estimation

packages/shared/src/schemas.ts
niklasnatter: yes you are right. tbh there isnt really a reason why i did that 😅 will add a validation for that case too

jerryafr (approved):

PR #1267, opened 2025-03-19
refactor(swap): inspect errors

acdibble:
the log will have `{ error: {} }`, instead of any useful information. if you use `inspect`, it will actually print the error


packages/swap/src/ingress-egress-tracking/index.ts
```diff
@@ -88,7 +89,7 @@ export const getPendingDeposit = async (
       txRef: deposits[0].tx_refs?.[0],
     };
   } catch (error) {
-    logger.error('error while looking up deposit in redis', { error });
+    logger.error('error while looking up deposit in redis', { error: inspect(error) });
```
niklasnatter, line 92: nice! i was assuming the previous way would be useful in the json logs - not sure if that was the case and this will break something?

niklasnatter (approved):

GabrielBuragev (approved):
nice, lets give it a try

PR #1268, opened 2025-03-20
chore(sdk): Pass full QuoteRequest object into API service

niklasnatter:


packages/sdk/src/swap/services/__tests__/__snapshots__/ApiService.test.ts.snap
```diff
@@ -82,6 +78,7 @@ exports[`ApiService > getQuote > gets a quote for a vault swap 2`] = `
 exports[`ApiService > getQuote > gets a quote with a broker commission 1`] = `
 {
   "amount": "10000",
+  "brokerCommissionBps": 15,
```
niklasnatter, line 81: value returned from service includes `brokerCommissionBps` now

jerryafr (approved):

PR #1271, opened 2025-03-24
feat(swap): handle on chain swap in swap requested [WEB-1962]

acdibble:


jerryafr (approved):

tinkaa01 (approved):

packages/swap/src/event-handlers/__tests__/__snapshots__/batchSwap.test.ts.snap
```diff
@@ -14,9 +14,9 @@ exports[`batch swap flow > handles all the events 1`] = `
   "estimatedExpiryAt": 1970-01-01T00:17:24.000Z,
   "expectedDepositAmount": null,
   "expiryBlock": null,
-  "fokMinPriceX128": null,
-  "fokRefundAddress": null,
-  "fokRetryDurationBlocks": null,
+  "fokMinPriceX128": "0",
```
sukhoviianastasia, line 17: Why there's no schema change on these params /swap/prisma.schema? or i missed it?

acdibble: They already existed. They became mandatory in 1.8 but we didn't generate the new event handlers until now.

PR #1272, opened 2025-03-25
chore: Move sdk path alias to package tsconfig

niklasnatter:


tsconfig.json
```diff
@@ -7,7 +7,6 @@
     "paths": {
       "@/shared/*": ["./packages/shared/src/*"],
       "@/swap/*": ["./packages/swap/src/*"],
-      "@/sdk/*": ["./packages/sdk/src/*"],
```
niklasnatter, line 10: we dont want other packages than the examples (eg the `swap` package) to import files from the sdk

niklasnatter: (this is a follow up for https://github.com/chainflip-io/chainflip-sdk-monorepo/pull/1274)

acdibble (approved):

PR #1274, opened 2025-03-25
chore: Remove swap server alias from tsconfig

niklasnatter:


packages/swap/src/event-handlers/__tests__/networkDepositFinalised.test.ts
```diff
@@ -1,7 +1,7 @@
 import { vi, describe, it, expect, beforeEach } from 'vitest';
-import { networkDepositFinalised } from '@/swap/event-handlers/networkDepositFinalised';
```
niklasnatter, line 2: the presence of `@/swap` in the `tsconfig.json` added an additional way to import files insied of the `swap` package and made out imports inconsistent between files. i dont think this is a problem, but it triggered me lol

acdibble (approved):

PR #1278, opened 2025-03-27
fix: Set typescript target to ES2024 to transform using keyword

niklasnatter:


packages/swap/Dockerfile
```diff
@@ -33,5 +33,4 @@ COPY --from=builder /app/packages/swap /app/packages/swap
 
 WORKDIR /app/packages/swap
 
-# using ENTRYPOINT causes signals (SIGINT) to be passed onto the node process
-ENTRYPOINT ["node", "--enable-source-maps", "./dist/swap/src/index.js"]
+CMD ["pnpm", "--filter", "swap", "start"]
```
niklasnatter, line 36: use start script like we do in web services

acdibble: we need this for k8s to be able to gracefully shut down the service
`# using ENTRYPOINT causes signals (SIGINT) to be passed onto the node process`

niklasnatter: changed

acdibble (approved):

PR #1281, opened 2025-03-28
fix(swap): use parsed query instead of raw query params

GabrielBuragev:


acdibble (approved):

packages/swap/src/routes/v2/quote.ts
```diff
@@ -91,7 +91,7 @@ export const validateQuoteQuery = async (query: Query) => {
     throw ServiceError.badRequest(amountResult.reason);
   }
 
-  if (query.isVaultSwap && query.srcChain === 'Polkadot') {
+  if (parsedQuery.isVaultSwap && assetConstants[parsedQuery.srcAsset].chain === 'Polkadot') {
```
sukhoviianastasia, line 94: ooops :/

sukhoviianastasia (approved):

PR #1283, opened 2025-04-01
fix(swap): handle padded versions [WEB-2154]

acdibble:


packages/shared/package.json
```diff
@@ -21,7 +21,7 @@
     "@chainflip/bitcoin": "^1.2.4",
     "@chainflip/rpc": "^1.8.9",
     "@chainflip/solana": "^2.1.4",
-    "@chainflip/utils": "^0.5.0",
+    "@chainflip/utils": "^0.8.0",
```
GabrielBuragev, line 24: we can probably move the @chainflip packages to the root package.json ?

GabrielBuragev (approved):

packages/shared/package.json
acdibble: we can't here because we publish packages and they need to all be listed in the `package.json` as dependencies

PR #1284, opened 2025-04-01
fix(swap): Handle brokerCommissionBps sent from SDK when encoding swap data

niklasnatter:


packages/sdk/src/swap/sdk.ts
```diff
@@ -499,7 +499,7 @@ export class SwapSDK {
       dcaParams: quote.type === 'DCA' ? quote.dcaParams : undefined,
       extraParams,
       brokerAccount,
-      brokerCommissionBps,
+      commissionBps: brokerCommissionBps,
```
niklasnatter, line 502: send `commissionBps` instead of `brokerCommissionBps` to backend (i did not want to rename in the zod schema because we use it in other places)

packages/swap/src/handlers/__tests__/__snapshots__/encodeVaultSwapData.test.ts.snap
```diff
@@ -2,16 +2,6 @@
 
 exports[`encodeVaultSwapData > rejects sanctioned addresses 1`] = `
 [
-  [
```
niklasnatter, line 5: mocks were not reset between tests so this included calls from other tests

acdibble (approved):

GabrielBuragev (approved):

PR #1287, opened 2025-04-03
feat: add assethub support

acdibble:


packages/shared/src/schemas.ts
```diff
@@ -146,10 +151,12 @@ export type FillOrKillParamsWithMinPrice = Omit<FillOrKillParamsX128, 'minPriceX
   minPrice: string;
 };
 
+export type AssetSymbol = (typeof assetConstants)[ChainflipAsset]['rpcAsset'];
```
GabrielBuragev, line 154: we can also just use `RpcAsset` from `@chainflip/utils` here or is that maybe ambiguous?

acdibble: it's not exported, i can do a new release with the export added

GabrielBuragev: Ah true, i just noticed that. Its also incomplete (doesn't have the hub assets), so it would be nice to add them and release anyway. 🙏

acdibble: all the assethub assets are there, `DOT`, `USDC`, and `USDT`

GabrielBuragev: indeed they are, excusez-moi

packages/shared/src/contracts.ts
```diff
@@ -1,7 +1,7 @@
+import { ChainflipNetwork } from '@chainflip/bitcoin';
```
GabrielBuragev, line 1: should import from `@chainflip/utils`

GabrielBuragev: we can then clean up the bitcoin package to use the exports from utils too

packages/shared/src/evm.ts
```diff
@@ -1,21 +1,22 @@
...
+export const assertIsEvmChain = (chain: ChainflipChain) => {
+  const evmChainId = getEvmChainId(chain, 'backspin');
   assert(evmChainId, `Chain ${chain} is not an evm chain`);
 };
 
-export const assertIsCCMDestination = (chain: Chain) => {
+export const assertIsCCMDestination = (chain: ChainflipChain) => {
   if (chain === 'Solana') return;
   assertIsEvmChain(chain);
 };
```
GabrielBuragev, line 15: I think these 2 assertions are not used anywhere?

GabrielBuragev: actually all of the assertions in this file

acdibble: yeah probably a ton of dead ccm logic, will double check

GabrielBuragev: Would be nice to 💣  this

acdibble: yeeted

packages/swap/src/event-handlers/__tests__/utils.ts
```diff
@@ -586,7 +514,7 @@ export const buildDepositFailedEvent = <T extends DepositFailedArgs>(args: T) =>
 };
 
 export const createChainTrackingInfo = (date?: Date) => {
-  const chains: Chain[] = ['Bitcoin', 'Ethereum', 'Polkadot', 'Solana'];
+  const chains: ChainflipChain[] = ['Bitcoin', 'Ethereum', 'Polkadot', 'Solana'];
   return Promise.all(
     chains.map((chain) =>
       prisma.chainTracking.upsert({
```
GabrielBuragev, line 520:
```suggestion
  return Promise.all(
    chainflipChains.map((chain) =>
      prisma.chainTracking.upsert({
```

packages/swap/src/ingress-egress-tracking/__tests__/index.test.ts
```diff
@@ -51,7 +52,7 @@ describe('ingress-egress-tracking', () => {
         'deposit:Ethereum:0x1234',
         JSON.stringify({
           amount: '0x9000',
-          asset: 'FLIP',
+          asset: { asset: 'FLIP', chain: 'Ethereum' },
```
GabrielBuragev, line 55: Not sure why this was needed here. Does it mean that we also have to adjust the deposit monitor tool to insert the data in this format?

packages/swap/src/quoting/Quoter.ts
```diff
@@ -1,16 +1,16 @@
...
 import { Subject, Subscription } from 'rxjs';
 import { Server, Socket } from 'socket.io';
-import {
-  InternalAsset,
-  InternalAssetMap,
-  UncheckedAssetAndChain,
-  assetConstants,
-  getInternalAsset,
-} from '@/shared/enums';
+import { InternalAssetMap } from '@/shared/enums';
```
GabrielBuragev, line 13: after releasing this - https://github.com/chainflip-io/chainflip-product-toolkit/pull/380 - we can use it from there

packages/swap/src/quoting/authenticate.ts
```diff
@@ -1,8 +1,9 @@
+import { ChainflipAsset, chainflipAssets, getInternalAsset } from '@chainflip/utils/chainflip';
 import * as crypto from 'crypto';
 import type { Server } from 'socket.io';
 import { promisify } from 'util';
 import { z } from 'zod';
-import { InternalAssetMap, getInternalAsset, InternalAsset, InternalAssets } from '@/shared/enums';
+import { InternalAssetMap } from '@/shared/enums';
```
GabrielBuragev, line 6: same here

GabrielBuragev (approved):
Lets release the latest utils and replace all `AssetSymbol`, `ChainMap`, `InternalAssetMap`, `AssetOfChain` occurances.

packages/swap/src/ingress-egress-tracking/__tests__/index.test.ts
acdibble: No this is the "new" format. It's been this way for a long time, but we never removed the old logic

packages/swap/src/quoting/Quoter.ts
acdibble: done

GabrielBuragev: we can probably do the same for AssetSymbol, ChainMap, InternalAssetMap, AssetOfChain

acdibble: entire file is yeeted

GabrielBuragev: there are some leftovers in the `schema.ts` file

acdibble: fixed that, just need to make sure now that the asset constants that we export match the shape of the old ones

PR #1308, opened 2025-05-05
feat(swap): handle solana fallback

acdibble:


packages/swap/src/backfillFallbacks.ts
```diff
@@ -0,0 +1,219 @@
...
+          asset: swapRequest.destAsset,
+        },
+      },
+    },
+  });
+};
+
+const solanaTransferFallbackQuery = gql(/* GraphQL */ `
+  query FallbackQuery {
+    events: allEvents(condition: { name: "SolanaIngressEgress.TransferFallbackRequested" }) {
```
jerryafr, line 188: maybe sorting by block ascending would be better

packages/swap/src/backfillFallbacks.ts
```diff
@@ -0,0 +1,219 @@
...
+          }
+        }
+      }
+    }
+  }
+`);
+
+const backfillFallbacks = async () => {
+  // finds every transfer fallback requested event for Solana
+  const result = await request(env.INGEST_GATEWAY_URL, solanaTransferFallbackQuery);
```
jerryafr, line 211: don't we want to backfill this for other chains? e.g. ethereum

packages/swap/src/event-handlers/ingress-egress/transferFallbackRequested.ts
```diff
@@ -0,0 +1,118 @@
...
+    const args = schemas[chain].parse(event.args);
+
+    if (!args.egressDetails) return;
+
+    const egresses = await prisma.broadcast
+      .findUnique({ where: { nativeId_chain: { chain, nativeId: args.broadcastId } } })
+      .egresses({ include: { swapRequests: true } });
+
+    if (egresses?.length !== 1) {
+      logger.warn('incorrect numbr of egresses found for transferFallbackRequested', {
```
jerryafr, line 79:
```suggestion
      logger.warn('incorrect number of egresses found for transferFallbackRequested', {
```

jerryafr (approved):

packages/swap/src/backfillFallbacks.ts
acdibble: the id in the indexer is `$BLOCK_ID-$INDEX_IN_BLOCK-$HASH` so it's safe to sort this way

PR #1321, opened 2025-05-15
chore(swap): handle refund fee

GabrielBuragev:


packages/swap/src/event-handlers/swapping/refundEgressScheduled.ts
```diff
@@ -16,27 +17,55 @@ export default async function refundEgressScheduled({
...
+      where: { nativeId: swapRequestId },
+      include: { fees: { where: { type: 'INGRESS' } } },
+    }),
+  ]);
+
+  const refundFee = swapRequest
+    .depositAmount!.minus(aggregate._sum.swapInputAmount ?? 0)
+    .minus(egressAmount?.toString() ?? 0)
+    .minus(egressFee?.toString() ?? 0)
+    .minus(swapRequest.fees.reduce((acc, fee) => acc.plus(fee.amount), new Prisma.Decimal(0)));
```
jerryafr, line 38: I think it's possible that swapRequest doesn't have ingress fees, right? would that cause issues here?

jerryafr: or this one is only for egress scheduled, if we've egress, then we definitely have had ingress?

GabrielBuragev: I think it shouldn't cause any issues as it will probably be set to 0 and `egressAmount` will just be higher.

jerryafr (approved):

PR #1324, opened 2025-05-20
feat(swap): add index prices to quote

acdibble:


GabrielBuragev (approved):
What do we want to do with this data?

packages/swap/prisma/schema.prisma
```diff
@@ -490,10 +490,14 @@ model Quote {
   quotedIntermediateAmount            Decimal?           @db.Decimal(30, 0)
   quotedEgressAmount                  Decimal            @db.Decimal(30, 0)
   quotedPrice                         Decimal
+  inputAssetPriceAtChannelOpening     Decimal?
```
niklasnatter, line 493: i think storing the `indexPriceAtChannelOpening`/`indexPriceAtCompletion`  could be convenient bc you can compare that one directly with `quotedPrice`/`executedPrice`

niklasnatter (approved):

packages/swap/prisma/schema.prisma
acdibble: yeah i'll add it

PR #1325, opened 2025-05-21
feat(swap): remove liquidity fees

acdibble:


packages/swap/src/routes/__tests__/quote.test.ts
```diff
@@ -386,19 +386,9 @@ describe('server', () => {
...
-              amount: '200000',
-              asset: 'USDC',
-              chain: 'Ethereum',
-            },
-            quoteAsset: {
-              asset: 'USDC',
-              chain: 'Ethereum',
-            },
+            baseAsset: { asset: 'ETH', chain: 'Ethereum' },
+            fee: { amount: '0', asset: 'USDC', chain: 'Ethereum' },
```
niklasnatter, line 390: could deprecate the `fee` property on the type

niklasnatter: or even the whole `poolInfo` property

niklasnatter (approved):

zoheb391:
nice. hopefully it fixes this without us needing to tell them again. 


Screenshot 2025-05-22 at 11 55 03

zoheb391 (approved):

packages/swap/src/routes/__tests__/quote.test.ts
acdibble: was considering it, not sure if we want to continue showing range order fees but that requires an enhancement to the rpc, so i'll keep it as is for now without deprecating

PR #1326, opened 2025-05-22
feat(swap): quote on chain swaps [WEB-2242]

acdibble:


packages/swap/src/routes/__tests__/v2/__snapshots__/quote.test.ts.snap
```diff
@@ -1,5 +1,207 @@
...
+    "recommendedSlippageTolerancePercent": 0.5,
+    "srcAsset": {
+      "asset": "USDC",
+      "chain": "Solana",
+    },
+    "type": "REGULAR",
+  },
+]
+`;
+
```
zoheb391, line 204: can we get a test for onChainSwaps with DCA? Asking because i don't think it will adjust duration according to chunks.

acdibble: changed one of the tests, everything works as expected

packages/swap/src/utils/getPoolQuote.ts
```diff
@@ -90,14 +106,24 @@ export default async function getPoolQuote({
...
+  includedFees.push({ ...ingressFee, type: 'INGRESS' });
+
+  if (isOnChain && networkFee.amount > env.MINIMUM_NETWORK_FEE_USDC) {
+    const networkFeeBps = isStableCoinSwap(srcAsset, destAsset)
+      ? env.ON_CHAIN_STABLECOIN_NETWORK_FEE_BPS
+      : env.ON_CHAIN_DEFAULT_NETWORK_FEE_BPS;
+    const normalNetworkFeeBps = 10n;
+    networkFee.amount = bigintMax(
+      (networkFee.amount * networkFeeBps) / normalNetworkFeeBps,
+      env.MINIMUM_NETWORK_FEE_USDC,
```
zoheb391, line 118: why are we dividing here? 🧐

zoheb391: also, the output is unaffected right? we gonna wait for protocol to change swap_rate_v3 ?

packages/swap/src/utils/getPoolQuote.ts
```diff
@@ -41,14 +48,23 @@ export default async function getPoolQuote({
...
     cfRateInputAmount -= boostFee;
   }
 
   if (isVaultSwap) {
     excludeFees.push('IngressDepositChannel');
   }
 
+  if (isOnChain) {
+    excludeFees.push('Egress');
+    excludeFees.push('IngressDepositChannel');
```
zoheb391, line 65: do we need to default brokerCommission to 0 for `onChainSwaps` ?

packages/swap/src/routes/__tests__/v2/__snapshots__/quote.test.ts.snap
zoheb391: 🙏

packages/swap/src/utils/getPoolQuote.ts
acdibble: added a check to the schema validation

niklasnatter, line 65: i think you also need to include the other ingress fee, otherwise the quote will include an ingress fee for vault swaps

niklasnatter: iirc bitcoin vault swaps have an egress fee because we stil need to fetch them

acdibble: so we should exclude this fee in most cases? we're not doing that currently

packages/swap/src/utils/getPoolQuote.ts
zoheb391: i meant as in the egress amount has not been adjusted to account for the new network fees?

packages/swap/src/utils/getPoolQuote.ts
niklasnatter: https://github.com/chainflip-io/chainflip-backend/blob/d5babf740948eccdaa9e1a281649a67c254af8fd/state-chain/runtime/src/lib.rs#L1799

acdibble: :)

zoheb391 (approved):

PR #1329, opened 2025-05-22
feat(sdk): add method to get extrinsic args [WEB-2243]

acdibble:


packages/sdk/src/swap/sdk.ts
```diff
@@ -426,6 +428,43 @@ export class SwapSDK {
     return this.trpc.encodeVaultSwapData.mutate(vaultSwapRequest);
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  getOnChainExtrinsicArgs({
```
niklasnatter, line 432:
```suggestion
  getOnChainSwapExtrinsicArgs({
```

niklasnatter (approved):

PR #1337, opened 2025-05-26
chore(swap): add indexes to swap, swapFee and swapRequest tables

jerryafr:
index hit rate was pretty low on these 3 tables and nativeId and swapRequestId are the fields that are used in where conditions
image


acdibble (approved):
nice, maybe update the description to include the query you used

PR #1347, opened 2025-06-04
feat(swap): add replenishment factor

acdibble:


packages/swap/src/config/env.ts
```diff
@@ -92,6 +92,11 @@ export default z
       z.number(),
     ),
     QUOTING_BASE_SLIPPAGE: internalAssetMap('QUOTING_BASE_SLIPPAGE', {}, z.number()),
+    QUOTING_REPLENISHMENT_FACTOR: internalAssetMap(
+      'QUTOING_REPLENISHMENT_MULTIPLIER',
```
CumpsD, line 96: typo `QUTOING_REPLENISHMENT_MULTIPLIER`

PR #1351, opened 2025-06-06
refactor(swap): restructure quoting

acdibble:


packages/swap/src/utils/getPoolQuote.ts
```diff
@@ -1,184 +0,0 @@
-import { internalAssetToRpcAsset } from '@chainflip/utils/chainflip';
```
niklasnatter, line 1: rip

packages/swap/src/quoting/QuoteRequest.ts
```diff
@@ -0,0 +1,487 @@
...
+    dcaParams,
+  }: { boostFeeBps?: number; dcaParams?: DcaParams } = {}) {
+    const includedFees = [];
+    const excludeFees: SwapFeeType[] = [];
+    let cfRateInputAmount = this.depositAmount;
+
+    // After this ticket, boost fee should be included in the response so dont have to calculate it ourselves
+    // https://linear.app/chainflip/issue/PRO-1370/include-boost-fees-in-quote-from-cf-swap-rate-v2
+    if (boostFeeBps) {
+      const boostFee = getPipAmountFromAmount(this.depositAmount, boostFeeBps);
```
zoheb391, line 225: would be great if we can remove this condition

zoheb391 (approved):
nice. much needed 👍

packages/swap/src/quoting/QuoteRequest.ts
```diff
@@ -0,0 +1,487 @@
...
+  }
+
+  private async setLimitOrders() {
+    if (this.limitOrders) return this.limitOrders;
+    this.limitOrders = await this.quoter.getLimitOrders(
+      this.srcAsset,
+      this.destAsset,
+      this.depositAmount,
+    );
+    return this.limitOrders;
```
niklasnatter, line 185: think we dont return for the other `set*` methods and we dont use the result

packages/swap/src/quoting/QuoteRequest.ts
acdibble: nice let me see

PR #1353, opened 2025-06-11
refactor(swap): return fees with amount 0

acdibble:


niklasnatter (approved):
do we want to though?

PR #1361, opened 2025-06-19
chore(swap): remove screening

GabrielBuragev:


packages/swap/src/handlers/__tests__/openSwapDepositChannel.test.ts
```diff
@@ -328,8 +323,6 @@ describe(openSwapDepositChannel, () => {
   });
 
   it('rejects sanctioned addresses', async () => {
```
zoheb391, line 325: why aren\t we removing these tests entirely?

acdibble (approved):

PR #1365, opened 2025-06-20
feat(sdk): update quoting to pass internal swap flag to rpc[ ]

sukhoviianastasia:


packages/swap/src/quoting/__tests__/QuoteRequest.test.ts
```diff
@@ -1,11 +1,65 @@
...
+  };
+});
+
+vi.mock('../../polkadot/api', async (importOriginal) => {
+  const original = (await importOriginal()) as object;
+  return {
+    ...original,
+    getInternalSwapNetworkFeeInfo: vi.fn(),
+  };
+});
```
acdibble, line 61: most of these can probably use a simple mock:
```suggestion
vi.mock('../../polkadot/api');
```

acdibble (approved):

PR #1377, opened 2025-07-01
feat(swap): add per lp replenishment factors [WEB-2359]

acdibble:


packages/swap/src/quoting/authenticate.ts
```diff
@@ -101,14 +99,39 @@ const authenticate = async (socket: QuotingSocket, next: Next) => {
...
+            acc[toLowerCase(factor.side)][factor.asset] =
+              factor.factor * (factor.side === 'BUY' ? 1 : -1);
+          }
+          return acc;
+        },
+        { buy: createInternalAssetMap(0), sell: createInternalAssetMap(0) },
+      ),
+      replenishmentFactors: marketMaker.factors.reduce(
+        (acc, factor) => {
+          if (factor.type === 'REPLENISHMENT') {
```
GabrielBuragev, line 118: we could perhaps filter them outside of the reducer, for both the `replenishmentFactors` and `mevFactors`

packages/swap/src/quoting/authenticate.ts
```diff
@@ -101,14 +99,39 @@ const authenticate = async (socket: QuotingSocket, next: Next) => {
...
+            if (factor.factor < 0) {
+              logger.error('replenishment factor is invalid', { factor });
+              return acc;
+            }
+            if (auth.quoted_assets[factor.asset]) {
+              acc[factor.asset] = factor.factor;
+            } else {
+              logger.warn('replenishment factor set for unquoted asset', {
```
GabrielBuragev, line 126: `attempted to set replenishment factor for unquoted asset`

packages/swap/src/quoting/Quoter.ts
```diff
@@ -347,4 +353,21 @@ export default class Quoter {
 
     return orders;
   }
+
+  private updateReplenishmentFactors() {
+    this.replenishmentFactors = createInternalAssetMap((asset) => {
+      const factor = this.accountIdToSocket
+        .values()
+        .reduce(
+          (acc, socket) => acc.plus(socket.data.replenishmentFactors[asset] ?? 0),
```
niklasnatter, line 362: i think the `replenishmentFactor` should only affect the liquidity of the account connected to the socket, no?

PR #1381, opened 2025-07-07
feat(swap): include boost delay blocks in swap duration calculation [WEB-2299]

GabrielBuragev:


packages/swap/src/event-handlers/ingress-egress/palletConfigUpdated.ts
```diff
@@ -0,0 +1,43 @@
...
+export type PalletConfigUpdatedArgsMap = {
+  [C in Chain]: z.input<(typeof palletConfigUpdatedSchemas)[C]>;
+};
+
+export const palletConfigUpdated =
+  (chain: Chain) =>
+  async ({ prisma, event }: EventHandlerArgs) => {
+    const { update } = palletConfigUpdatedSchemas[chain].parse(event.args);
+
+    if (update.__kind.startsWith('SetBoostDelay') && 'delayBlocks' in update) {
```
acdibble, line 30: can define a helper to match and then you don't have to use `in`

```ts
const enumMatches = <E extends { __kind: string }, const T extends string>(
  e: E,
  prefix: T,
): e is Extract<E, { __kind: `${T}${string}` }> => e.__kind.startsWith(prefix);
```

packages/swap/src/utils/boost.ts
```diff
@@ -40,3 +41,11 @@ export const getBoostFeeBpsForAmount = async ({
     maxBoostFeeBps,
   };
 };
+
+export const getBoostChainflipBlocksDelayForChain = async (chain: Chain): Promise<number> => {
+  const boostDelay = await prisma.boostDelayChainflipBlocks.findFirst({
```
acdibble, line 46: we usually read these values from an RPC to ensure they are the latest, because the processor has a delay. here the impact is minimal if we have some stale data but still think an RPC would be preferable. seems like this value should be added to the ingress/egress environment RPC?

acdibble (approved):

packages/swap/src/utils/boost.ts
GabrielBuragev: However, i still see this being used on `main`

PR #1399, opened 2025-08-08
feat(swap): handle cancel swaps

zoheb391:
closes WEB-2511


GabrielBuragev (approved):
I guess it'll make sense to wait for the event renaming on the protocol before merging this, but looks good otherwise

PR #1423, opened 2025-09-25
chore: remove polkadot support

zoheb391:
closes WEB-2642

### DIFF
--- a/packages/swap/python-client/quoting_client.py
+++ b/packages/swap/python-client/quoting_client.py
@@ -27,7 +27,6 @@ cf_assets = [
     ("Ethereum", ["ETH", "USDT", "USDC", "FLIP"]),
     ("Arbitrum", ["ETH", "USDC"]),
     ("Bitcoin", ["BTC"]),
-    ("Polkadot", ["DOT"]),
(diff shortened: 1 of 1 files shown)


acdibble (approved):

packages/swap/src/quoting/Leg.ts
```diff
@@ -11,22 +12,23 @@ export default class Leg {
...
+    private readonly destAsset: Exclude<ChainflipAsset, 'Dot'>,
     public amount: bigint,
   ) {}
 
   private getSide(): 'BUY' | 'SELL' {
     return this.destAsset !== 'Usdc' ? 'BUY' : 'SELL';
   }
 
-  getBaseAsset(): Exclude<ChainflipAsset, 'Usdc'> {
+  getBaseAsset(): Exclude<BaseChainflipAsset, 'Dot'> {
```
GabrielBuragev, line 31: I think we need to exclude both `Usdc` & `Dot` here

zoheb391: baseChainflipASset excludes usdc already

PR #1438, opened 2025-10-01
fix(swap): handle new error

acdibble:


zoheb391 (approved):

jerryafr (approved):
maybe we can throw different insufficient liquidity errors so that we can distinguish later if needed.
this error can potentially raise because of other issues too 🤔

PR #1448, opened 2025-11-03
feat(swap): update event handlers

acdibble:


packages/swap/src/event-handlers/swapping/swapRequested.ts
```diff
@@ -40,6 +40,10 @@ const getRequestInfo = (requestType: RequestType) => {
       };
     }
 
+    if (requestType.outputAction.__kind === 'CreditLendingPool') {
+      return { type: 'INTERNAL' as const, destAddress: undefined, ccmMetadata: undefined };
+    }
+
```
zoheb391: should we add lending todos here?

packages/swap/src/event-handlers/ingress-egress/depositFinalised.ts
```diff
@@ -1,48 +1,48 @@
...
 }));
-const polkadotSchema = polkadotSchema190.transform((args) => ({
+const polkadotSchema = polkadotSchema11200.transform((args) => ({
   ...args,
   depositDetails: { chain: 'Polkadot' as const, data: args.depositDetails },
   depositAddress:
     args.depositAddress && formatForeignChainAddress({ __kind: 'Dot', value: args.depositAddress }),
 }));
-const solanaSchema = solanaSchema190.transform((args) => ({
+const solanaSchema = solanaSchema11200.transform((args) => ({
```
zoheb391: do we need union here?

PR #1549, opened 2026-04-09
chore(swap): check only JIT Lps for undeployed liquidity

zoheb391:
closes WEB-3252


jerryafr (approved):
instead of removing the previous approach, can we add a feature flag where we can switch between these 2 incase something goes wrong?

PR #1552, opened 2026-04-15
feat(swap, sdk): add broker id to query params to /v2/quote

sukhoviianastasia:


packages/swap/src/quoting/QuoteRequest.ts
```diff
@@ -160,6 +181,10 @@ export default class QuoteRequest {
     this.pools = params.pools;
     this.ccmParams = params.ccmParams;
     this.brokerCommissionBps = params.brokerCommissionBps;
+    this.integrator = params.integrator;
+    this.broker = params.broker;
+    this.userAgent = params.userAgent;
+    this.sdkVersion = params.sdkVersion;
```
sukhoviianastasia: userAgent and sdkVersion might not be necessary

sukhoviianastasia:
i have `brokerIdSs58` in `swap` package and `brokerAccount` in `sdk`. the reason is that `VaultSwapRequest ` in `sdk` already had `brokerAccount` and although they have different functions, conceptually they are same

Quote url will look like this: 
```
https://chainflip-swap.chainflip.io/v2/quote
  ?amount=50000000
  &srcChain=Bitcoin&srcAsset=BTC
  &destChain=Ethereum&destAsset=USDC
  &isVaultSwap=false&isOnChain=false
  &brokerCommissionBps=15
  &dcaEnabled=true&dcaV2Enabled=true
  &broker=cFLdocJo3bjT7JbT7R46cA89QfvoitrKr9P3TsMcdkVWeeVLa
``` 
i think integrators will need to be notified to: 
    * Bump SDK and pass { integrator, brokerAccount } to new SwapSDK(...), or
    * Append the two query params directly if they hit the REST endpoint without the SDK.

packages/shared/src/schemas.ts
```diff
@@ -114,6 +116,15 @@ export const quoteQuerySchema = z
...
     }
 
+    if (args.brokerIdSs58 !== undefined && !(args.brokerIdSs58 in brokerAliasMap)) {
+      ctx.addIssue({
+        message: `unknown broker account: ${args.brokerIdSs58}`,
+        code: z.ZodIssueCode.custom,
+      });
+
+      hadError = true;
+    }
```
jerryafr: we should not do this check. otherwise we'll have to keep the map up to date all the time which is impossible

jerryafr (approved):

PR #1556, opened 2026-04-17
feat(swap): publish messages for specific quoter orders

jerryafr:


packages/swap/src/quoting/Quoter.ts
```diff
@@ -96,7 +101,10 @@ export type QuotingSocket = Socket<ReceivedEventMap, SentEventMap, any, SocketDa
 export default class Quoter {
   private readonly quotes$ = new Subject<Quote>();
 
-  private readonly inflightRequests = new Set<string>();
+  private readonly inflightRequests = new Map<
+    string,
+    { quoteRequestId: string | undefined; responded: Set<AccountId> }
```
zoheb391, line 106:
```suggestion
    { quoteRequestId: string | undefined; responders: Set<AccountId> }
```

i would expect responded to be a boolean

packages/swap/src/messageQueues/quoteEvents.ts
```diff
@@ -5,24 +5,18 @@ import baseLogger, { logStorage } from '../utils/logger.js';
...
+    if (v === null || v === undefined) return v;
+    if (Array.isArray(v)) return v.map(serialize);
+    if (typeof v === 'object') {
+      return Object.fromEntries(
+        Object.entries(v as Record<string, unknown>).map(([k, val]) => [k, serialize(val)]),
+      );
+    }
+    return v;
+  };
+  return serialize(obj) as Record<string, unknown>;
```
sukhoviianastasia, line 19: how did you realize that you need to refactor it? do we need to move it to utils?

zoheb391 (approved):

packages/swap/src/quoting/Quoter.ts
```diff
@@ -306,6 +358,15 @@ export default class Quoter {
...
             });
+            publishQuoteOrderError({
+              quoteRequestId: logStorage.getStore(),
+              marketMaker: accountId,
+              marketMakerRequestId: requestId,
+              error: 'insufficient balance',
+              balance,
+              amount,
+              sellAsset,
+            });
```
sukhoviianastasia, line 369: so i can see that we publish errors in these 3 cases: 
 - `insufficient balance`
 - `tick provided is too small`
 - `unknown request_id`

but there are more errors in Quoter, fe:
- `unexpected missing format function` (response was parsed but wasn't asked to quote this request)
- mismatched request_id (line 299)
- market maker disconnected while request is in flight (line 144)

or we don't need to handle these error situations?

sukhoviianastasia, line 369: also this specific `publishQuoteOrderError` call seems to be very nested (legIndex → for accountId → for [tick, amount]), it's going to result pushing many messages; but idk if there's anything we can do here

packages/swap/src/messageQueues/quoteEvents.ts
jerryafr: it's a simpler implementation. already had some tests for it.
https://github.com/chainflip-io/chainflip-sdk-monorepo/blob/8ca69e42fbd12f36ee0494833a01c69c82d13906/packages/swap/src/messageQueues/__tests__/quoteEvents.test.ts#L148

packages/swap/src/quoting/Quoter.ts
jerryafr: if those errors are catched, we will publish them.
i've allowed the `error` field to be any string to make it as flexible as possible. this way if we introduce other error types in the quoter, no changes in backoffice will be needed.
in backoffice we can aggregate with error types and display whatever we receive from quoter

jerryafr: you're right. we maybe be able to improve it by grouping errors together and publish one large message. but i'd choose that if we face issues with the current flow

sukhoviianastasia (approved):

PR #1571, opened 2026-05-07
feat(swap): add a rate limiter for quoting endpoint

jerryafr:


packages/swap/src/routes/quoteRateLimit.ts
```diff
@@ -0,0 +1,77 @@
...
+  timestamps: number[];
+  blockedUntil: number;
+};
+
+export const createQuoteRateLimit = (): RequestHandler => {
+  if (!env.ENABLE_QUOTE_RATE_LIMIT) {
+    return (_req, _res, next) => next();
+  }
+
+  const records = new Map<string, IpRecord>();
```
sukhoviianastasia, line 18: there might be many ips, no? but then we are only cleaning un-blocked ips?

packages/swap/src/routes/quoteRateLimit.ts
```diff
@@ -0,0 +1,77 @@
...
+      return next(ServiceError.tooManyRequests());
+    }
+
+    if (record.blockedUntil !== 0) {
+      record.timestamps = [];
+      record.blockedUntil = 0;
+    }
+
+    const cutoff = now - windowMs;
+    record.timestamps = record.timestamps.filter((t) => t >= cutoff);
```
sukhoviianastasia: filter will create new array each time, can use `splice` instead, but it's a minor thing probably

sukhoviianastasia (approved):

packages/swap/src/routes/quoteRateLimit.ts
jerryafr: we're cleaning them up as well in that interval

PR #1577, opened 2026-05-13
feat(sdk): include btc supply pool in getBoostLiquidity method [WEB-3162]

GabrielBuragev:


packages/sdk/src/swap/sdk.ts
```diff
@@ -278,27 +284,49 @@ export class SwapSDK {
...
+          /** @deprecated DEPRECATED(2.2): Only value supported moving forward is 5 bps */
+          feeTierBps?: number;
+        }
+      | ({
+          /** @deprecated DEPRECATED(2.2): Only value supported moving forward is 5 bps */
+          feeTierBps?: number;
+        } & UncheckedAssetAndChain) = {},
   ): Promise<BoostPoolDepth[]> {
-    let poolsDepth = await this.getBoostPoolsDepth();
+    if (params.feeTierBps && params.feeTierBps !== 5) {
```
sukhoviianastasia, line 297: will not throw if `params.feeTierBps` is 0 but that probably won't happen (coz people used to pass 5, 10 and 30)

GabrielBuragev: Hmm, yeah technically we shouldn't return any data if they request the liquidity at 0 bps anyways, as 0 is the equivalent to "no boost" 😅

sukhoviianastasia (approved):

PR #1583, opened 2026-05-19
feat(sdk, example): fix tron vault swap

zoheb391:
closes WEB-3418

Screenshot 2026-05-19 at 17 42 56

### DIFF
--- a/packages/examples/src/tronVaultSwap.ts
+++ b/packages/examples/src/tronVaultSwap.ts
@@ -81,7 +81,7 @@ if (sourceTokenAddress) {
     { feeLimit: 100_000_000 },
     [
       { type: 'address', value: to },
-      { type: 'uint256', value: BigInt(swapAmount).toString() },
+      { type: 'uint256', value: swapAmount },
     ],
     tronAddress as string,
   );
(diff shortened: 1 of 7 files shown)


packages/examples/src/tronVaultSwap.ts
```diff
@@ -5,107 +5,107 @@ import { TronWeb } from 'tronweb';
...
-  process.env.TRON_SECRET_KEY = account.privateKey;
-
-  console.log('TRON_SECRET_KEY not set - generating random secret key');
-  console.log('generated secret key', process.env.TRON_SECRET_KEY);
+  // TronWeb expects private keys without the 0x prefix
+  process.env.TRON_SECRET_KEY = account.privateKey.replace(/^0x/, '');
+  console.log('TRON_SECRET_KEY not set - generated random secret key', process.env.TRON_SECRET_KEY);
+  throw new Error(
+    'you need to fund this account otherwise TronWeb will pretend like it sent funds without actually sending anything',
+  );
```
GabrielBuragev: we can probably also just assert that a `TRON_SECRET_KEY` has been passed over as an env var, otherwise throw.

zoheb391: claude gave a good suggestion. to check if tx is successful so we don't need this anymore

packages/examples/src/tronVaultSwap.ts
```diff
@@ -5,107 +5,108 @@ import { TronWeb } from 'tronweb';
...
-          { name: '_value', type: 'uint256' },
-        ],
-        outputs: [{ name: '', type: 'bool' }],
-      },
+      { type: 'address', value: to },
+      { type: 'uint256', value: BigInt(swapAmount).toString() },
     ],
-    sourceTokenAddress,
+    tronAddress as string,
+  );
```
GabrielBuragev, line 87: Should this be `value` instead, similar to how it is in the else block?

```suggestion
      { type: 'uint256', value: BigInt(value).toString() },
```

GabrielBuragev (approved):

PR #1584, opened 2026-05-20
feat(swap): skip publishing order events for beta LPs

jerryafr:


GabrielBuragev (approved):

zoheb391 (approved):

packages/swap/src/quoting/Quoter.ts
sukhoviianastasia, line 396: why not here?

jerryafr: on line 368, we already make sure the LP is not beta. beta ones will never get to this publish.
```tsc
if (quote.beta) continue;
```

sukhoviianastasia: thanks!!!

PR #1595, opened 2026-06-12
fix(swap): effective boost fee bps rounding errors

GabrielBuragev:


zoheb391 (approved):
maybe a comment describing why we need the migration (even in the PR is fine)

packages/swap/prisma/migrations/20260612125436_update_effective_boost_fee_bps_4_to_5/migration.sql
```diff
@@ -0,0 +1,4 @@
+-- Data migration: bump effectiveBoostFeeBps from 4 to 5 on existing SwapRequest rows
+UPDATE "public"."SwapRequest"
+SET "effectiveBoostFeeBps" = 5
+WHERE "effectiveBoostFeeBps" = 4;
```
GabrielBuragev, line 4: Due to rounding errors caused by the divisions with big numbers, we ended up storing incorrect values (off by 1 unit) for these columns. Minimum boost fee bps was always 5 bps, as the smallest fee tier pool was the one with 5 bps. Moving forward, we end up with one 5 bps boost pool and one 5 bps supply pool.

PR #1626, opened 2026-08-10
chore: remove indexer gateway dependency

zoheb391:
this allows us to yeet the indexer gateway permanently


packages/swap/src/indexer.ts
```diff
@@ -0,0 +1,73 @@
+import pg from 'pg';
```
jerryafr, line 1: we could use prisma here 🤔 
no strong opinion though

jerryafr (approved):

GabrielBuragev (approved):

packages/swap/src/indexer.ts
zoheb391: would avoid so we don't need another schema file.

PR #1635, opened 2026-08-20
fix(sdk): tolerate brokerAccount and brokerIdSs58 [WEB-3882]

sukhoviianastasia:


CHANGELOG.md
```diff
@@ -7,6 +7,22 @@ intended to be used with.
...
+  param, matching the `brokerAccount` option on `getQuoteV2` and the field name
+  used by the deposit channel and vault swap endpoints. Integrators calling the
+  REST API directly should send `brokerAccount`; SDK users keep passing
+  `brokerAccount` to `getQuoteV2` (or `SwapSDKOptions`) and need no changes.
+
+### Deprecated
+
+- The `brokerIdSs58` query param on `/v2/quote` is deprecated in favour of
+  `brokerAccount`. It is still accepted, since SDK versions 2.2.0 - 2.3.x send
+  the broker account under that name. If both are set, `brokerAccount` wins.
```
sukhoviianastasia: or should I say that we accept both params for REST API? help me communicate it correctly, please

packages/shared/src/schemas.ts
```diff
@@ -141,7 +142,7 @@ export const quoteQuerySchema = z
       dcaV2Enabled: args.dcaV2Enabled,
       isVaultSwap: args.isVaultSwap,
       isOnChain: args.isOnChain,
-      brokerIdSs58: args.brokerIdSs58,
+      brokerIdSs58: args.brokerAccount ?? args.brokerIdSs58,
```
sukhoviianastasia: i only tolerate both params for REST API. should i do same for SDK? SDK accepts `brokerAccount` only

GabrielBuragev: I'd make `args.brokerIdSs58` take precedence.
```suggestion
      brokerIdSs58: args.brokerIdSs58 ??args.brokerAccount,
```

sukhoviianastasia: so guess leave it as it is then

sukhoviianastasia: test was also adjusted to "prefers `brokerIdSs58` when both names are set"

GabrielBuragev (approved):

PR #1638, opened 2026-09-11
feat(swap,sdk): [WEB-3805] add requestId to quote response headers and sdk

jerryafr:


packages/sdk/src/swap/services/ApiService.ts
```diff
@@ -40,14 +40,16 @@ export const getQuoteV2: BackendQuery<
...
     baseURL: baseUrl,
     params,
     signal,
     headers: CF_SDK_VERSION_HEADERS,
   });
 
-  return { ...quoteRequest, quotes: data };
+  const quoteRequestId = headers?.['x-request-id'] as string | undefined;
+
+  return { ...quoteRequest, quotes: data, ...(quoteRequestId && { quoteRequestId }) };
```
GabrielBuragev: should we maybe simply call it `requestId` ?

GabrielBuragev (approved):
Nice, i think we can simply call it `requestId` so it follows the same convention in every other endpoint (if needed). Also the header is called `x-request-id` too so it follows that naming convention too.

PR #1639, opened 2026-09-17
chore: add 2.3 changelog & add some more vault swap examples

GabrielBuragev:
Closes WEB-3832


sukhoviianastasia (approved):

zoheb391:
did you try running the deprecations script to see if we get any failures?

GabrielBuragev:
> did you try running the deprecations script to see if we get any failures?

I did check the code manually. I'll run the script now however, good point

GabrielBuragev:
> did you try running the deprecations script to see if we get any failures?

No deprecations in effect found for this version. Also just fyi, the script is a part of the CI process so we should get alerted by a failing pipeline. Thanks again for the heads up

